PR #2, opened 2017-10-28
clientconfig package

jtopjian:
This commit introduces the clientconfig package. clientconfig
is based on the Python os-client-config library and will offer
convenience functions for creating OpenStack clients.

### DIFF
--- a/acceptance/openstack/clientconfig/clientconfig_test.go
+++ b/acceptance/openstack/clientconfig/clientconfig_test.go
@@ -1,0 +1,39 @@
+// +build acceptance clientconfig
+
+package clientconfig
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+
+	acc_compute "github.com/gophercloud/gophercloud/acceptance/openstack/compute/v2"
+	acc_tools "github.com/gophercloud/gophercloud/acceptance/tools"
+
+	cc "github.com/gophercloud/utils/openstack/clientconfig"
+)
+
+func TestServerCreateDestroy(t *testing.T) {
+	clientOpts := &cc.ClientOpts{
+		Cloud:     "acctest",
+		EnvPrefix: "FOO",
+	}
+
+	client, err := cc.NewServiceClient("compute", clientOpts)
+	if err != nil {
+		t.Fatalf("Unable to create client: %v", err)
+	}
+
+	server, err := acc_compute.CreateServer(t, client)
+	if err != nil {
+		t.Fatalf("Unable to create server: %v", err)
+	}
+	defer acc_compute.DeleteServer(t, client, server)
+
+	newServer, err := servers.Get(client, server.ID).Extract()
+	if err != nil {
+		t.Fatalf("Unable to get server %s: %v", server.ID, err)
+	}
+
+	acc_tools.PrintResource(t, newServer)
+}
--- a/openstack/auth_yaml.go
+++ b/openstack/auth_yaml.go
@@ -1,0 +1,131 @@
+package openstack
+
+import (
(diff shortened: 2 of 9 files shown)


jtopjian:
@jrperritt This PR started as an attempt to port https://github.com/gophercloud/gophercloud/pull/44. My intention was to make a single simple function that read and parsed a `clouds.yaml` file. But in order to do that correctly, it made sense to try to make a Go/Gophercloud version of `os-client-config`.

I've attempted to recreated a small portion of it here. I left out areas which aren't applicable to Gophercloud (either at all or not yet supported).

This is pretty much the bare minimum required to do anything useful with clouds.yaml+gophercloud.

IMO, this has a lot of potential. It can enhance Gophercloud usage by abstracting certain nuances with client creation. From the opposite direction, it can handle things like SSL options, providing a single way doing that for all/most applications.

Let me know what you think.

jtopjian:
Thanks! It's got enough functionality to create a client. I'm going to begin integrating it into Terraform. There might be a few follow-up patches required.

Question: The `findAndReadYAML` function is utilizing the `github.com/Wessie/appdirs` package. This package detects system and user config directories across Mac, Linux, and Windows, so there's a good amount of stuff going on to warrant not implementing the work locally. Should this be vendored?

jrperritt:
Thanks for pointing that out; I'd missed it. I'll take a look

jrperritt:
OK, I just took another look at the bit that's using the `appdirs` package. How do feel about forgoing that bit for now, possibly adding it later? Though I understand searching all these "special" directories may be convenient for many users, I'm not too keen on magic like that. Namely, my preference (as a user) is to provide the directory I want it to search (though I recognize that OpenStack will search the current directory, `~/.config/openstack`, and `/etc/openstack` by default).

jtopjian:
I am in full agreement with not searching special directories (or any hard-coded directory) for something that goes in `gophercloud/gophercloud`. 

The Python `os-client-config` package searches all of the directories specified in this PR, which is why I included it here (In fact, the choice of `github.com/Wessie/appdirs` was because its advertised as a clone of the Python `appdirs` which `os-client-config` uses).

I'm personally not a fan of `appdirs`, either. Even if I was using a Windows workstation, I'd do whatever was possible to have a unix-style configuration structure. But that's me and people who end up using an application which consumes this might be different.

How about a compromise of dropping `appdirs` for now, but still searching the directories which do not require `appdirs`? If there's enough feedback to include the full suite of directories, we can look at adding it back. 

If there are users in the meantime who are left out, they can still set the `OS_CLIENT_CONFIG_FILE` environment variable to the specific full path to the file.

Another thing to keep in mind is that `os-client-config` has the ability to merge/nest configuration files to a certain extent. This is not implemented here (yet. maybe never), but that might be one reason why so many directories are searched. I can see an organization which works on several clouds have staff use workstations which are maintained by a central IT support department. Provisioning of those workstations includes deployment of a base `clouds.yaml` in one of the `appdirs` directories.

I'm only imagining that scenario as a possible reason why there are so many directories. I feel like, at least right now, the chances of that scenario being used for `gophercloud/utils` is quite improbable. So it's quite safe to leave out for now.

jrperritt:
> How about a compromise of dropping appdirs for now, but still searching the directories which do not require appdirs?

Yes, I think we agree, I just didn't explain myself well enough: searching current directory, ~/.config/openstack, and /etc/openstack is fine. I'm only suggesting removing the `appdirs` bit.

jtopjian:
@jrperritt Ping

jtopjian:
This should be squashed, but I want to make sure @raphink's initial work makes it into the git history.

raphink:
Thanks for the intention @jtopjian

PR #3, opened 2017-11-13
Use user_domain_* instead of domain_*

mcanevet:


jtopjian:
@mcanevet Thank you for pointing this out. Rather than replacing `DomainID` with `UserDomainID` (and `*Name`), `UserDomainID` and `UserDomainName` need to be added as additional fields. We then need to determine the correct `DomainID` and `DomainName` setting by following the precedence shown here:

https://github.com/openstack/os-client-config/blob/master/os_client_config/config.py#L593-L609

I've opened issue #4 to track this:

https://github.com/gophercloud/utils/issues/4

If you'd like to work on this, you're more than welcome to. Otherwise, I can take a look in the next day or two.

mcanevet:
@jtopjian where would you put the code that determines the correct `DomainID` and `DomainName` to use?

jtopjian:
It could go in the `AuthOptions` function here:

https://github.com/gophercloud/utils/blob/master/openstack/clientconfig/requests.go#L104-L111

After re-reading the Python function again, I think it should be fine to add something like this:

```go
	if auth.UserDomainID != "" {
		ao.DomainID = auth.UserDomainID
	}

	if auth.UserDomainName != "" {
		ao.DomainName = auth.UserDomainName
	}
```

I think this might cause some scoping issues in certain situations, but I think that would need to be dealt with in Gophercloud (https://github.com/gophercloud/gophercloud/issues/252).

mcanevet:
@jtopjian PR amended.

jtopjian:
@mcanevet looks good, but the unit tests aren't passing (we need to get Travis hooked up to this repo).

The Domain settings need to be updated in the `testing/fixtures.go` file.

mcanevet:
@jtopjian unit tests fixed

jtopjian:
@mcanevet I think something might have gone awry with the last push. I see less code than before :)

mcanevet:
@jtopjian I think it's normal, I removed the changes I did on unit tests because they are not required anymore with your suggestion to add fields and set the correct `DomainID` and `DomainName`.

Issue #5, opened 2017-11-14
Travis Support


jtopjian:
We should integrate Travis testing into this repo.

jtopjian:
@jrperritt Hate to push this one on you, but it'll need an auth token.

jrperritt:
No problem, I'll take care of it

jrperritt:
I've enabled the Travis CI service for this repo, but it isn't yet in a position to run the same travis.yml file as in `gophercloud/gophercloud` (e.g. no `script` directory).

@jtopjian If this bug is for adding the travis.yml file as well, I'm not sure when I'll get to that part.

jtopjian:
@jrperritt Would you be able to commit a basic `.travis.yml` file with the [`secure`](https://github.com/gophercloud/gophercloud/blob/master/.travis.yml#L14) token? I can work on getting actual tests to run after that.

jrperritt:
Maybe? It looks like I added that, but I don't remember what's encrypted in that token, nor how I encrypted it.

jrperritt:
I don't have time to look into it in depth, but after briefly reading some of the Travis CI docs, that token may not be necessary.

jtopjian:
Yep, looks like you're right. The token must be for some non-basic features like releases and such.

jtopjian:
I think we're good to go for now. Thank you!

PR #6, opened 2017-11-16
initial travis support

jtopjian:
For #5


jtopjian:
@jrperritt Ping for review.

Here's an example of a failed run: https://travis-ci.org/gophercloud/utils/builds/302817871

PR #9, opened 2017-12-31
Implement Gnocchi archive policy list

ozerovandrei:
Add Gnocchi archive policy structure and list request.
Add unit tests and fake Gnocchi service client for tests.

The same command with Gnocchi CLI is done with:
`gnocchi archive-policy list`

For [#698](https://github.com/gophercloud/gophercloud/issues/698)

Links to the line numbers/files in the Gnocchi source code that support the
code in this PR:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L333
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/indexer/sqlalchemy.py#L563
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/archive_policy.py#L29

PR #10, opened 2017-12-31
Implement Gnocchi service client

ozerovandrei:
Add NewGnocchiV1 to work with Gnocchi v1 REST API.

code in this PR:
https://github.com/gnocchixyz/python-gnocchiclient/blob/stable/4.0/gnocchiclient/v1/client.py#L41

### DIFF
--- a/gnocchi/client.go
+++ b/gnocchi/client.go
@@ -1,0 +1,25 @@
+package gnocchi
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+func initClientOpts(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts, clientType string) (*gophercloud.ServiceClient, error) {
+	sc := new(gophercloud.ServiceClient)
+	eo.ApplyDefaults(clientType)
+	url, err := client.EndpointLocator(eo)
+	if err != nil {
+		return sc, err
+	}
+	sc.ProviderClient = client
+	sc.Endpoint = url
+	sc.Type = clientType
+	return sc, nil
+}
+
+// NewGnocchiV1 creates a ServiceClient that may be used with the v1 Gnocchi package.
+func NewGnocchiV1(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
+	sc, err := initClientOpts(client, eo, "metric")
+	sc.ResourceBase = sc.Endpoint + "v1/"
+	return sc, err
+}


jtopjian:
@ozerovandrei Thank you for working on this!

I think we should get some basic acceptance test files in place here, even if they're not used yet. In the main Gophercloud repo, check out `acceptance/clients/clients.go`. At a minimum, let's implement a `GnocchiV1Client` function. That way we can write a very brief snippet to validate that a client can successfully connect to the Gnocchi service without error. This is to ensure the client works before moving on to service communication.

Let me know if you have any questions.

ozerovandrei:
@jtopjian I guess that you mean NewGnocchiV1Client? Check [10c9fc8](https://github.com/gophercloud/utils/pull/10/commits/10c9fc87c727da4e7d270c5891722b7090a51f03).

jtopjian:
@ozerovandrei Yes, thank you :smile: 

I'm sure this works, but I want to set up a local Gnocchi environment to verify everything. This might take me a few days to get to which I apologize for.

Additionally, we should be able to get automated acceptance tests working for Gnocchi, too, which will help validate things.

ozerovandrei:
@jtopjian for a local API tests [gnocchi-docker](https://github.com/gnocchixyz/gnocchi-docker) or [gnocchi-devstack](https://github.com/gnocchixyz/gnocchi/tree/master/devstack) should be enough.
 To test all of the Gnocchi REST stuff you really need only indexer for a metadata (PostgreSQL/MySQL) and a metric storage backend (local filesystem/Redis) (don't waste your time with CEPH for that).
And the Gnocchi-API service on top of that (you don't need the Gnocchi-Metricd service because you don't need to care about saving real measures to the backend).

It would be better to have the automated testing like the main Gophercloud repo has, of course. But I don't know how to help with that task.

jtopjian:
>  It  would be better to have the automated testing like the main Gophercloud  repo has, of course. But I don't know how to help with that task.

Yep, that task is in progress - nothing to do on your part :)

I don't think it'll be difficult at all to set up a local test environment, I just haven't found the time yet. This week has been busy for me. I've only had an hour or so a day which I've used reviewing other PRs. I apologize about that. I do hope to get this moving soon.

jtopjian:
@ozerovandrei Thank you for your patience. I was able to get a basic Gnocchi environment up and running tonight.

I did a quick test of the client + #9. I can see http activity between Gophercloud, Gnocchi, and Keystone so that's good enough for me for a start. If there are bug for some reason, we can patch them as we go.

PR #11, opened 2018-01-03
Implement Gnocchi metric list

ozerovandrei:
Add Gnocchi metric structure and list request.

`gnocchi metric list`

code in this PR:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L447
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/indexer/sqlalchemy.py#L689

### DIFF
--- a/acceptance/gnocchi/metric/v1/metrics_test.go
+++ b/acceptance/gnocchi/metric/v1/metrics_test.go
@@ -1,0 +1,31 @@
+package v1
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/utils/acceptance/clients"
+	"github.com/gophercloud/utils/gnocchi/metric/v1/metrics"
+)
+
+func TestMetricsList(t *testing.T) {
+	client, err := clients.NewGnocchiV1Client()
+	if err != nil {
+		t.Fatalf("Unable to create a Gnocchi client: %v", err)
+	}
+
+	listOpts := metrics.ListOpts{}
+	allPages, err := metrics.List(client, listOpts).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to list metrics: %v", err)
+	}
+
+	allMetrics, err := metrics.ExtractMetrics(allPages)
+	if err != nil {
+		t.Fatalf("Unable to extract metrics: %v", err)
+	}
+
+	for _, metric := range allMetrics {
+		tools.PrintResource(t, metric)
+	}
+}
--- a/acceptance/gnocchi/metric/v1/metrics_test.go
+++ b/acceptance/gnocchi/metric/v1/metrics_test.go
@@ -1,3 +1,5 @@
+// +build acceptance metric metrics
+
 package v1
 
 import (
--- a/gnocchi/metric/v1/metrics/doc.go
+++ b/gnocchi/metric/v1/metrics/doc.go
@@ -1,0 +1,24 @@
+/*
+Package metrics provides the ability to retrieve metrics through the Gnocchi API.
+
+Example of Listing metrics
+
+	listOpts := metrics.ListOpts{
+		Limit: 25,
+	}
+
+	allPages, err := metrics.List(gnocchiClient, listOpts).AllPages()
+	if err != nil {
+		panic(err)
+	}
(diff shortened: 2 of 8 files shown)


ozerovandrei:
Acceptance test will be provided after implementing Gnocchi service client:
#10 

Unit tests won't work until fake Gnocchi service client will be implemented in:
#9

ozerovandrei:
@jtopjian this is ready for a review.

jtopjian:
@ozerovandrei I think the [`MetricsController`](https://github.com/gnocchixyz/gnocchi/blob/df2b64026c6ea697c47200a9fb7ec2b02622de92/gnocchi/rest/api.py#L551) is the right controller here instead of the `MetricController`. The `/v1` URL directs to the [`V1Controller`](https://github.com/gnocchixyz/gnocchi/blob/df2b64026c6ea697c47200a9fb7ec2b02622de92/gnocchi/rest/api.py#L2130) and from there, [`/metric`](https://github.com/gnocchixyz/gnocchi/blob/df2b64026c6ea697c47200a9fb7ec2b02622de92/gnocchi/rest/api.py#L2141) goes to `MetricsController`. A little confusing...

Anyway, [list/get_all](https://github.com/gnocchixyz/gnocchi/blob/df2b64026c6ea697c47200a9fb7ec2b02622de92/gnocchi/rest/api.py#L666) for `MetricsController` looks to have some undocumented list parameters:

* `user_id`
* `project_id`
* `creator`

You can verify this by doing:

```bash
#!/bin/bash

oscurl() {                                     
  token=$(openstack token issue -c id -f value)
  curl -s -H "X-Auth-Token:$token" $1          
}                                              

oscurl http://localhost:8041/v1/metric?user_id=7db0ef67a7a52490e836ae1ba7f367395

oscurl http://localhost:8041/v1/metric?user_idddd=7db0ef67a7a52490e836ae1ba7f367395
```

Note the deliberate typo in the second `oscurl` command to make sure Gnocchi isn't accepting _anything_. As well, if you run the first command as a non-admin user, you will get a 403. But if you run it as an admin, it works. So we should add those parameters to `ListOpts`.

Next, we hit some policy checks: https://github.com/gnocchixyz/gnocchi/blob/df2b64026c6ea697c47200a9fb7ec2b02622de92/gnocchi/rest/api.py#L690-L696

And finally the actual retrieval of metrics: https://github.com/gnocchixyz/gnocchi/blob/df2b64026c6ea697c47200a9fb7ec2b02622de92/gnocchi/rest/api.py#L698-L708

which goes to the SQLAlchemy link you gave.

The actual model for a metric is here: https://github.com/gnocchixyz/gnocchi/blob/df2b64026c6ea697c47200a9fb7ec2b02622de92/gnocchi/indexer/sqlalchemy_base.py#L74

It looks like there's a field called [`archive_policy_name`](https://github.com/gnocchixyz/gnocchi/blob/df2b64026c6ea697c47200a9fb7ec2b02622de92/gnocchi/indexer/sqlalchemy_base.py#L119) have you come across this before? 

So in summary:

1. I think we should add the additional fields to `ListOpts`
2. Maybe add the `archive_policy_name` to the result? If you don't have a fixture for this, that's fine - leave it missing, but it looks like it's an actual field.

Thoughts?

Other than that, this looks really good to me and no other changes are needed -- nice work :)

ozerovandrei:
@jtopjian there is an additional feature when you creating Gnocchi resources - you can create it as usual or you can create a resource and a metric simultaneously in the one request. You need to pass the `archive_policy_name` into a such request: [resource-with-metrics](https://gnocchi.xyz/rest.html#with-metrics).

I think that we don't need the `archive_policy_name` inside a `Metric` result because it will be empty every time, and Gophercloud users should use the `Metric.ArchivePolicy.Name` to get needed information.

jtopjian:
What about this example: https://gnocchi.xyz/rest.html#create ?

I'm not concerned about the POST request right now, but the response will have `archive_policy_name` which means it will have to be added as a field to the `Metric` struct. It looks like you can either leave `archive_policy_name` out for now and add it when you implement "Create" or add it now - up to you :)

jtopjian:
LGTM! Nice work :)

PR #12, opened 2018-01-03
Implement Gnocchi resource list

ozerovandrei:
Add Gnocchi resource structure and list request.

`gnocchi resource list`

code in this PR:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L1126
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/indexer/sqlalchemy.py#L988

### DIFF
--- a/acceptance/gnocchi/metric/v1/resources_test.go
+++ b/acceptance/gnocchi/metric/v1/resources_test.go
@@ -1,3 +1,5 @@
+// +build acceptance metric resources
+
 package v1
 
 import (
--- a/acceptance/gnocchi/metric/v1/resources_test.go
+++ b/acceptance/gnocchi/metric/v1/resources_test.go
@@ -1,0 +1,32 @@
+package v1
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/utils/acceptance/clients"
+	"github.com/gophercloud/utils/gnocchi/metric/v1/resources"
+)
+
+func TestResourcesList(t *testing.T) {
+	client, err := clients.NewGnocchiV1Client()
+	if err != nil {
+		t.Fatalf("Unable to create a Gnocchi client: %v", err)
+	}
+
+	opts := resources.ListOpts{}
+	resourceType := ""
+	allPages, err := resources.List(client, opts, resourceType).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to list resources: %v", err)
+	}
+
+	allResources, err := resources.ExtractResources(allPages)
+	if err != nil {
+		t.Fatalf("Unable to extract resources: %v", err)
+	}
+
+	for _, resource := range allResources {
+		tools.PrintResource(t, resource)
+	}
+}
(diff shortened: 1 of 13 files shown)


ozerovandrei:
Unit tests won't work until fake Gnocchi service client will be implemented in: [#9](https://github.com/gophercloud/utils/pull/9)

Acceptance test will be provided after implementing Gnocchi service client: [#10](https://github.com/gophercloud/utils/pull/10)

jtopjian:
@ozerovandrei I did a quick review of this and wanted to add some notes. 

Looking at the code, it looks like there's an additional option of [`history`](https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L1129). 

It looks like you can specify but both `detail` and `history` in the URL _and_ a header:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L179-L191

I think focusing only on the URL parameters is fine for now.

The fields in the `results.go` file look correct, but the API docs seem to include extra fields depending on the resource type. For example, the resource type `instance` has a `display_name` field. To handle these extra fields in Gophercloud, we've used a field called `Extra` like this:

https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/users/results.go#L12-L77

jtopjian:
> Also I think that I should unmarshal RevisionStart, RevisionEnd, StartedAt, EndedAt into the JSONRFC3339Milli. Because Gnocchi always returns timestamps in that format: rest/timestamp-format.

That's a good idea :)

ozerovandrei:
@jtopjian for now I've done the custom `JSONRFC3339NanoTimezone` inside the `gnocchi/results.go` because Gnocchi responses have the following format: `2006-01-02T15:04:05.999999+00:00`.
Unfortunately built-in Gophercloud result types don't have such format. So I've done the new inside this repo.

jtopjian:
> for now I've done the custom JSONRFC3339NanoTimezone inside the gnocchi/results.go because Gnocchi responses have the following format: 2006-01-02T15:04:05.999999+00:00.
Unfortunately built-in Gophercloud result types don't have such format. So I've done the new inside this repo.

Looks good to me.

> Also I've looked to the history request again. That request needs a new url with Gnocchi resource id like in the usual Gnocchi resource get request (GET v1/resource/type/UUID/history). So that feature needs to be implemented inside another PR. I guess that issue for that PR will be created inside this repo sometime later.

That sounds good. Having additional functions beyond List, Get, Create, Update, and Delete is normal for some API endpoints.

> Also I have a problem. To implement unmarshaling to an Extra field I need to import github.com/gophercloud/gophercloud/internal. But Go doesn't allow such imports if the importing code is outside the tree rooted at the parent of the internal directory: proposal. So I need to duplicate internal inside utils repo.
> Can I do that? Or it's better to rename internal package inside the main repo via Issue+PR? It only used in gophercloud/openstack/identity/v3 for now.

Oh right - good point. The function is _very_ small, so it's easiest to duplicate it here. Go ahead and mirror the directory structure, files, and tests in this `utils` repo.

jtopjian:
@ozerovandrei This looks really good - very nice job on this.

PR #13, opened 2018-01-05
Implement Gnocchi resource get

ozerovandrei:
Add Gnocchi resource get request and unit test.

`gnocchi resource show`

code in this PR:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L994
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/indexer/sqlalchemy.py#L936

### DIFF
--- a/gnocchi/metric/v1/resources/urls.go
+++ b/gnocchi/metric/v1/resources/urls.go
@@ -5,16 +5,10 @@ import "github.com/gophercloud/gophercloud"
 const resourcePath = "resource"
 
 func rootURL(c *gophercloud.ServiceClient, resourceType string) string {
-	if resourceType == "" {
-		resourceType = "generic"
-	}
 	return c.ServiceURL(resourcePath, resourceType)
 }
 
 func resourceURL(c *gophercloud.ServiceClient, resourceType, resourceID string) string {
-	if resourceType == "" {
-		resourceType = "generic"
-	}
 	return c.ServiceURL(resourcePath, resourceType, resourceID)
 }
 
--- a/gnocchi/metric/v1/resources/urls.go
+++ b/gnocchi/metric/v1/resources/urls.go
@@ -11,13 +11,17 @@ func rootURL(c *gophercloud.ServiceClient, resourceType string) string {
 	return c.ServiceURL(resourcePath, resourceType)
 }
 
-func listURL(c *gophercloud.ServiceClient, resourceType string) string {
-	return rootURL(c, resourceType)
-}
-
-func getURL(c *gophercloud.ServiceClient, resourceID string, resourceType string) string {
+func resourceURL(c *gophercloud.ServiceClient, resourceType, resourceID string) string {
 	if resourceType == "" {
 		resourceType = "generic"
 	}
 	return c.ServiceURL(resourcePath, resourceType, resourceID)
+}
+
+func listURL(c *gophercloud.ServiceClient, resourceType string) string {
+	return rootURL(c, resourceType)
+}
+
+func getURL(c *gophercloud.ServiceClient, resourceType, resourceID string) string {
+	return resourceURL(c, resourceType, resourceID)
 }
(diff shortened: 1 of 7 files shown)


ozerovandrei:
This PR depends on [#12](https://github.com/gophercloud/utils/pull/12)

ozerovandrei:
@jtopjian this is ready but it needs to be rebased and merged after #12.

jtopjian:
I just merged #12 (I'm assuming it was ready for merge since you removed `[wip]`).

ozerovandrei:
@jtopjian yeah, that's right.

gnocchi/metric/v1/resources/urls.go
```diff
@@ -11,6 +11,17 @@ func rootURL(c *gophercloud.ServiceClient, resourceType string) string {
 	return c.ServiceURL(resourcePath, resourceType)
 }
 
+func resourceURL(c *gophercloud.ServiceClient, resourceType, resourceID string) string {
+	if resourceType == "" {
+		resourceType = "generic"
+	}
```
jtopjian: I think we should require the user to specify a resource type instead of providing a default of "generic". Having `""` as a parameter always looks strange.

Keep in mind that this package/library shouldn't be end-user facing, so the application which consumes this can always provide a default of "generic". Does that make sense? Thoughts?

ozerovandrei:
@jtopjian "generic" is a special type of a resource. All Gnocchi resources have this type and some of them additionally can have some other types (like "instance").

So for example when user has 3 resources with type "_instance_" and 2 with type "_instance_disk_" and makes a list request with type "_generic_" he gets a result with 5 resources. 
Same with get request. That user can create a get request with a "_generic_" resource type in the url to his "_instance_" resource and he will get a result (**BUT** he will see the **"type": "instance"** in the response, funny right?).

So he'll need to provide a default "generic" type or a valid specific type and both will work.

Given that, i think it's better to treat "generic" resources inside Gophercloud as a special default resource type.

jtopjian:
@ozerovandrei I think I get the logistics of it - thanks.

However, to confirm, the way you have `urls.go` configured right now, these two functions are the same:

```go
resource, err := resources.Get(gnocchiClient, "", resourceID).Extract()
resource, err := resources.Get(gnocchiClient, "generic", resourceID).Extract()
```

If this is true, the better design is to not use the second. Using `""` as an "alias" for `generic` isn't helping anyone -- it might actually confuse some developers.

Now, if I were to write a command-line tool called `gognocchi`, I can see having an optional command-line flag of `--resource-type` that defaults to "generic". But the communication between Go and the API should be explicit.

But maybe I'm still not understanding?

ozerovandrei:
@jtopjian yep you are right. I've just tried to implement some over-complicated logic that shouldn't be here. And also I've done some mistakes.

Right now we have functions `rootURL` (used for list/create and it's merged into master) and `resourceURL` (used for get/update) and they both tries to convert empty resource type to a _generic_.

If I make a get request to a `v1/resource/UUID` to read one resourse, I'll get the 404 error.
If I make a get request to a `v1/resource/generic/UUID` to read one resourse, I'll get that resource. And it doesn't matter if that resource have type `instance`. I also can do a get to `v1/resource/instance/UUID` (but I can't do the `v1/resource/disk/UUID`, for example).

If I make a get request to a `v1/resource` to read a list, I'll get a list of resource types with urls like that (well, that's interesting, haven't touched that 😄 ):
```
{
  "generic": "url/v1/resource/generic",
  "instance": "url/v1/resource/instance"
}
```
If I make a get request to a `v1/resource/generic` to read a list, I'll get all resources. And it doesn't matter if my resources have different types.
If I make a get request to a `v1/resource/instance` to read a list, I'll get resources with the `instance` type. 

So yeah, I've done completely confusing stuff in the `urls.go` and left users without any choice.
I'll fix both `rootURL` and `resourceURL` and all tests.

jtopjian:
@ozerovandrei No problem at all. Code review can feel like a negative thing since the reviewer is always pointing out mistakes, but you're doing great work with all of these PRs.

Once we start merging Create, Update, and Delete, we might find a few bugs missed in other PRs which is totally fine -- nobody is perfect :)

jtopjian:
@ozerovandrei I just want to confirm: this is ready to merge?

ozerovandrei:
@jtopjian yeah, it is ready now.

PR #16, opened 2018-01-15
Implement Gnocchi resource creation

ozerovandrei:
Add Gnocchi resource create request with tests and documentation.

`gnocchi resource create`

code in this PR:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L1092
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/indexer/sqlalchemy.py#L745

### DIFF
--- a/gnocchi/metric/v1/resources/doc.go
+++ b/gnocchi/metric/v1/resources/doc.go
@@ -35,8 +35,6 @@ Example of Creating a resource without a metric
 
 	createOpts := resources.CreateOpts{
 		ID: "23d5d3f7-9dfa-4f73-b72b-8b0b0063ec55",
-		ProjectID: "4154f088-8333-4e04-94c4-1155c33c0fc9",
-		UserID: "bd5874d6-6662-4b24-a9f01c128871e4ac",
 	}
 	resourceType = ""
 	resource, err := resources.Create(gnocchiClient, createOpts, resourceType).Extract()
@@ -49,7 +47,6 @@ Example of Creating a resource with links to some existing metrics
 	createOpts := resources.CreateOpts{
 		ID: "23d5d3f7-9dfa-4f73-b72b-8b0b0063ec55",
 		ProjectID: "4154f088-8333-4e04-94c4-1155c33c0fc9",
-		UserID: "bd5874d6-6662-4b24-a9f01c128871e4ac",
 		Metrics: map[string]interface{}{
 			"disk.read.bytes.rate": "ed1bb76f-6ccc-4ad2-994c-dbb19ddccbae",
 			"disk.write.bytes.rate": "0a2da84d-4753-43f5-a65f-0f8d44d2766c",
--- a/gnocchi/metric/v1/resources/doc.go
+++ b/gnocchi/metric/v1/resources/doc.go
@@ -42,13 +42,13 @@ Example of Creating a resource without a metric
 		panic(err)
 	}
 
-Example of Creating a resource with links to some existing metrics with an ending timestamp of the resource
+Example of Creating a resource with links to some existing metrics with a starting timestamp of the resource
 
-	endedAt := time.Date(2018, 1, 4, 10, 0, 0, 0, time.UTC)
+	startedAt := time.Date(2018, 1, 4, 10, 0, 0, 0, time.UTC)
 	createOpts := resources.CreateOpts{
 		ID: "23d5d3f7-9dfa-4f73-b72b-8b0b0063ec55",
 		ProjectID: "4154f088-8333-4e04-94c4-1155c33c0fc9",
-		EndedAt: &endedAt,
+		StartedAt: &startedAt,
 		Metrics: map[string]interface{}{
 			"disk.read.bytes.rate": "ed1bb76f-6ccc-4ad2-994c-dbb19ddccbae",
 			"disk.write.bytes.rate": "0a2da84d-4753-43f5-a65f-0f8d44d2766c",
--- a/gnocchi/metric/v1/resources/doc.go
+++ b/gnocchi/metric/v1/resources/doc.go
@@ -44,7 +44,7 @@ Example of Creating a resource without a metric
 
 Example of Creating a resource with links to some existing metrics with an ending timestamp of the resource
 
-	endedAt := time.Time(2018, 1, 10, 21, 0, 0, 0, time.UTC)
+	endedAt := time.Date(2018, 1, 4, 10, 0, 0, 0, time.UTC)
 	createOpts := resources.CreateOpts{
 		ID: "23d5d3f7-9dfa-4f73-b72b-8b0b0063ec55",
 		ProjectID: "4154f088-8333-4e04-94c4-1155c33c0fc9",
--- a/gnocchi/metric/v1/resources/doc.go
+++ b/gnocchi/metric/v1/resources/doc.go
@@ -36,7 +36,7 @@ Example of Creating a resource without a metric
 	createOpts := resources.CreateOpts{
 		ID: "23d5d3f7-9dfa-4f73-b72b-8b0b0063ec55",
 		ProjectID: "4154f088-8333-4e04-94c4-1155c33c0fc9",
-		UserID: "bd5874d666624b24a9f01c128871e4ac",
+		UserID: "bd5874d6-6662-4b24-a9f01c128871e4ac",
 	}
 	resourceType = ""
 	resource, err := resources.Create(gnocchiClient, createOpts, resourceType).Extract()
@@ -49,8 +49,8 @@ Example of Creating a resource with links to some existing metrics
 	createOpts := resources.CreateOpts{
 		ID: "23d5d3f7-9dfa-4f73-b72b-8b0b0063ec55",
 		ProjectID: "4154f088-8333-4e04-94c4-1155c33c0fc9",
-		UserID: "bd5874d666624b24a9f01c128871e4ac",
-		Metrics: map[string]string{
+		UserID: "bd5874d6-6662-4b24-a9f01c128871e4ac",
+		Metrics: map[string]interface{}{
 			"disk.read.bytes.rate": "ed1bb76f-6ccc-4ad2-994c-dbb19ddccbae",
 			"disk.write.bytes.rate": "0a2da84d-4753-43f5-a65f-0f8d44d2766c",
 		},
@@ -66,8 +66,8 @@ Example of Creating a resource and a metric a the same time
 	createOpts := resources.CreateOpts{
 		ID: "23d5d3f7-9dfa-4f73-b72b-8b0b0063ec55",
 		ProjectID: "4154f088-8333-4e04-94c4-1155c33c0fc9",
-		UserID: "bd5874d666624b24a9f01c128871e4ac",
-		Metrics: map[string]map[string]string{
+		UserID: "bd5874d6-6662-4b24-a9f01c128871e4ac",
+		Metrics: map[string]interface{}{
 			"cpu.delta": map[string]string{
 				"archive_policy_name": "medium",
 			},
--- a/gnocchi/metric/v1/resources/doc.go
+++ b/gnocchi/metric/v1/resources/doc.go
@@ -31,11 +31,10 @@ Example of Getting a resource
 		panic(err)
 	}
 
-Example of Creating a resource without a metric with a string timestamp for the starting time
+Example of Creating a resource without a metric
 
 	createOpts := resources.CreateOpts{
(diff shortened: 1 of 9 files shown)


ozerovandrei:
@jtopjian I have one question - is there any helper function to generate UUIDs in Gophercloud? Gnocchi wants and ID of the resource inside the creation request. For now I've just did static hardcoded UUID for the acceptance tests helper: `acceptance/gnocchi/metric/v1/resources.go`.

jtopjian:
@ozerovandrei Gophercloud has some basic "random" functions here: https://github.com/gophercloud/gophercloud/blob/master/acceptance/tools/tools.go

If Gnocchi needs a UUID specifically, you can use https://github.com/satori/go.uuid for the acceptance tests.

ozerovandrei:
@jtopjian it seems that Travis doesn't want to cooperate. But this PR is ready.

jtopjian:
@ozerovandrei Yeah, I think something is wrong with Travis in general.

This looks good - nice work with the documentation as well as the thorough unit tests. I find it odd that Gnocchi isn't auto-generating a UUID itself, but oh well.

I think there are two additional things we should do here:

1. When creating a non-generic resource, it looks like Gnocchi is accepting one or more arbitrary key/value pairs (for example, `image_id` for the "instance" resource). We can handle this similar to the "Extra" field in results, but in the opposite direction.

There are a few areas in Gophercloud that do this. Off of the top of my head, look at how `Extra` is handled [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/users/requests.go).

I think calling the field "Extra" is fine unless you think something like "ExtraAttributes" or "AdditionalAttributes" sounds better (in which case we should also rename the field in `results.go`).

2. The `StartedAt` and `EndedAt` fields should probably be a proper `time.Time` field. There isn't precedence for this in Gophercloud so I can't link to an example, but doing some kind of time-to-string conversion in the `ToResourceCreateMap` function should work fine. Those fields should only be added to the request body if they are not a zero-value (Or make them `*time.Time` so you can easily check if the value is `nil` or not).

Let me know if that sounds incorrect, if you have any questions, or need any help.

gnocchi/metric/v1/resources/requests.go
```diff
@@ -99,11 +102,19 @@ type CreateOpts struct {
...
 	UserID string `json:"user_id,omitempty"`
 
-	// StartedAt is a resource creation timestamp.
-	StartedAt string `json:"started_at,omitempty"`
+	// StartedAtTimeStamp is a resource creation timestamp. It's contents will be converted into the string
+	// and then transferred to the private "startedAt" field.
+	StartedAtTimeStamp *time.Time `json:"-"`
+
+	// StartedAtString is a string representation of the "StartedAt" field.
+	StartedAtString string `json:"started_at,omitempty"`
```
jrperritt: Let's drop these `String` fields; just have `StartedAt` (no `TimeStamp`). Same with `EndedAt`.

gnocchi/metric/v1/resources/requests.go
```diff
@@ -99,11 +102,19 @@ type CreateOpts struct {
 	// UserID is the Identity user of the resource.
 	UserID string `json:"user_id,omitempty"`
 
-	// StartedAt is a resource creation timestamp.
-	StartedAt string `json:"started_at,omitempty"`
+	// StartedAtTimeStamp is a resource creation timestamp. It's contents will be converted into the string
```
jrperritt: "It's" is short for "It is". Here we want "Its" (no apostrophe).

gnocchi/metric/v1/resources/requests.go
ozerovandrei: @jrperritt @jtopjian I have a problem with that. If I'm providing a time.Time or a &time.Time into Gophercloud's `func BuildRequestBody`, it returns error at the [L156](https://github.com/gophercloud/gophercloud/blob/master/params.go#L156):
```
2018/01/17 22:43:55 json: cannot unmarshal string into Go value of type map[string]interface {}
```
It's printing a string representation of a time.Time at [L152](https://github.com/gophercloud/gophercloud/blob/master/params.go#L152) if it's uncommented:
```
string(b): "2018-01-02T23:23:34Z"
```

Can you give an advice what is better to do?
Or I should create an issue inside Gophercloud's main repo and we will talk about it inside and leave that PR for now?

jtopjian: @ozerovandrei Try calling `b, err := gophercloud.BuildRequestBody(opts, "")` at the beginning of the function and then work on all of the special/unique fields.

ozerovandrei: @jtopjian that's the thing. When I'm providing an opts and some filed in the opts has the `time.Time` or the `&time.Time` type, the `BuildRequestBody` throws an error that is shown in my previous comment.

jtopjian: @ozerovandrei Oops - I'm really sorry about that. 

You're right: there's a bug in `gophercloud/gophercloud/params.go`. 

I see what you were trying to accomplish with the `StartedAtString` fields now. Instead of using those fields, the following also works:

```go
func (opts CreateOpts) ToResourceCreateMap() (map[string]interface{}, error) {
  var startedAt, endedAt string

  if opts.StartedAt != nil {
    startedAt = opts.StartedAt.Format(gnocchi.RFC3339NanoTimezone)
    opts.StartedAt = nil
  }

  if opts.EndedAt != nil {
    endedAt = opts.EndedAt.Format(gnocchi.RFC3339NanoTimezone)
    opts.EndedAt = nil
  }

  b, err := gophercloud.BuildRequestBody(opts, "")
  if err != nil {
    return nil, err
  }

  if startedAt != "" {
    b["started_at"] = startedAt
  }

  if endedAt != "" {
    b["ended_at"] = endedAt
  }

  if opts.ExtraAttributes != nil {
    for key, value := range opts.ExtraAttributes {
      b[key] = value
    }
  }

  return b, nil
}
```

_but_ that's a hack and I think we should fix `params.go` instead.

@jrperritt The `time.Time` field is getting passed back in `BuildRequestBody` [here](https://github.com/gophercloud/gophercloud/blob/master/params.go#L138). What's the best way to handle this? This feels just as hacky as the above:

```go
if v.Type().String() != "*time.Time" {
  //fmt.Printf("Calling BuildRequestBody with:\n\tv: %+v\n\tf.Name:%s\n", v.Interface(), f.Name)
  _, err := BuildRequestBody(v.Interface(), f.Name)
  if err != nil {
    return nil, err
  }
}
```

jrperritt: @jtopjian I think in `BuildRequestBody`, while iterating the struct fields, we can check toward the top of the loop if the `json` struct tag is "-" and `continue` if it is

jtopjian: @jrperritt Thank you! 😄 

@ozerovandrei This should now be fixed in https://github.com/gophercloud/gophercloud/pull/730.  The `ToResourceCreateMap` should now be able to work as:

```go
// ToResourceCreateMap constructs a request body from CreateOpts.
func (opts CreateOpts) ToResourceCreateMap() (map[string]interface{}, error) {
  b, err := gophercloud.BuildRequestBody(opts, "")
  if err != nil {
    return nil, err
  }

  if opts.StartedAt != nil {
    b["started_at"] = opts.StartedAt.Format(gnocchi.RFC3339NanoTimezone)
  }

  if opts.EndedAt != nil {
    b["ended_at"] = opts.EndedAt.Format(gnocchi.RFC3339NanoTimezone)
  }

  if opts.ExtraAttributes != nil {
    for key, value := range opts.ExtraAttributes {
      b[key] = value
    }
  }

  return b, nil
}
```

ozerovandrei: @jtopjian thank you!

jtopjian:
@ozerovandrei Is this ready for review?

ozerovandrei:
@jtopjian yeah. All opened PR's for Gnocchi are ready.

jtopjian:
This looks good to me.

ozerovandrei:
@jtopjian I rechecked that in my installation with the latest `4.1.3` release and yeah, you right. Gnocchi  doesn't allow me to create a resource with any of `created_by_project_id `, `created_by_user_id `, `creator` fields.
Values for those field are populated automatically after resource creation on the server side.

So, to tune some Identity info of a resource `project_id`, `user_id` can be provided in the create request. Same with the update request.

I will fix my PRs.

jtopjian:
> So, to tune some Identity info of a resource project_id, user_id can be provided in the create request. Same with the update request.

I have a feeling that the reason `project_id` and `user_id` are accepted is because Gnocchi is accepting almost anything (ie: `image_id` for an "instance"). It's probably best to double-check the Gnocchi code and confirm what the base/core fields for Create and Update are.

ozerovandrei:
@jtopjian thank you for such analysis.

`Gnocchi is accepting almost anything (ie: image_id for an "instance")`
Well it's a half-true. It depends on a [resource type](https://gnocchi.xyz/rest.html#resource-types) of a resource. So, for example, if the resource type `instance` has the attribute `image_id`, then you can create a resource with a type `instance` and provide a value for that attribute.
You can't create a generic resource with that field:
```
curl v1/resource/generic -d '{"id": "00000000-0000-dead-beef-111111111111", "image_id": "00000000-0000-dead-beef-111111111112"}'

The server could not comply with the request since it is either malformed or otherwise incorrect.<br /><br />
Invalid input: extra keys not allowed @ data[u'image_id']
```

`I think we should remove original_resource_id, too.`
That's right, I am sorry that I didn't test all of those parameters at the beginning. Gnocchi doesn't allow `original_resource_id` to be provided in the create or update request too.

So yeah, 
```
the following are the right fields for Create:

ID
Metrics
ProjectID
UserID
StartedAt
EndedAt
ExtraAttributes
```

those are fields that can be provided in the create and update requests.

ozerovandrei:
@jtopjian @jrperritt all PRs about Gnocchi resources were fixed and rebased.

jtopjian:
> Well it's a half-true. It depends on a resource type of a resource. So, for example, if the resource type instance has the attribute image_id, then you can create a resource with a type instance and provide a value for that attribute.

Right. I had it backwards, actually. Depending on the resource type, Gnocchi will only accept the valid fields for that type. 

> That's right, I am sorry that I didn't test all of those parameters at the beginning. Gnocchi doesn't allow original_resource_id to be provided in the create or update request too.

It's no problem at all. Putting this stuff together is by no means an easy task - there's a lot of investigative work and things slip by.

gnocchi/metric/v1/resources/doc.go
```diff
@@ -31,5 +31,51 @@ Example of Getting a resource
...
+	createOpts := resources.CreateOpts{
+		ID: "23d5d3f7-9dfa-4f73-b72b-8b0b0063ec55",
+	}
+	resourceType = "generic"
+	resource, err := resources.Create(gnocchiClient, resourceType, createOpts).Extract()
+	if err != nil {
+		panic(err)
+	}
+
+Example of Creating a resource with links to some existing metrics with an ending timestamp of the resource
```
jtopjian: I think this might be a bug with Gnocchi itself.

If no start date is specified, Gnocchi uses the current time.

If both a start and end date are given, Gnocchi will ensure the start date is earlier than the end date. 

But if only an end date is specified, Gnocchi will still use the current time, which means the start date will come _after_ the end date.

Again, I think this is a bug in Gnocchi. But because of it, we probably should not use it as an example :)

ozerovandrei:
@jtopjian I've fixed that.

I think that the best time to add a creation of non-generic resources is when I'll implement a support for resource types API in the future for a separate issue.

jtopjian:
> I think that the best time to add a creation of non-generic resources is when I (or who knows, maybe someone else) will implement a support for resource types API in the future for a separate issue.

That's fine. The important thing is to prepare the acceptance tests for that kind of functionality now instead of doing a lot of refactoring later.

This looks good to me. For real this time :)

I'll merge once @jrperritt has done another review.

ozerovandrei:
@jrperritt can you please recheck that?

jtopjian:
@jrperritt quick ping on this one, too.

@ozerovandrei The use of the `required` tag in #22 made me double-check this PR. It looks like `ID` is the only required field. Rather than adding it here, though, I'm OK if you want to add it in #17 so you don't have to do a bunch of rebasing.

jtopjian:
@jrperritt Thank you!

PR #17, opened 2018-01-16
Implement Gnocchi resource update

ozerovandrei:
Add Gnocchi resource create request with tests and documentation.

The same command with Gnocchi CLI is done with:
`gnocchi resource update`

For [#698](https://github.com/gophercloud/gophercloud/issues/698)

Links to the line numbers/files in the OpenStack source code that support the
code in this PR:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L1007
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/indexer/sqlalchemy.py#L788

PR #18, opened 2018-01-16
Implement Gnocchi resource delete

ozerovandrei:
Add Gnocchi resource delete request with tests and documentation.

`gnocchi resource delete`

code in this PR:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L1050
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/indexer/sqlalchemy.py#L896

### DIFF
--- a/acceptance/gnocchi/metric/v1/resources.go
+++ b/acceptance/gnocchi/metric/v1/resources.go
@@ -40,3 +40,16 @@ func CreateGenericResource(t *testing.T, client *gophercloud.ServiceClient) (*re
 	t.Logf("Successfully created the generic Gnocchi resource.")
 	return resource, nil
 }
+
+// DeleteResource will delete a Gnocchi resource with specified type and ID.
+// A fatal error will occur if the delete was not successful.
+func DeleteResource(t *testing.T, client *gophercloud.ServiceClient, resourceType, resourceID string) {
+	t.Logf("Attempting to delete the Gnocchi resource: %s", resourceID)
+
+	err := resources.Delete(client, resourceType, resourceID).ExtractErr()
+	if err != nil {
+		t.Fatalf("Unable to delete the Gnocchi resource %s: %v", resourceID, err)
+	}
+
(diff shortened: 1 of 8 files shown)


ozerovandrei:
@jtopjian I've provided a custom header into the `Delete` method's `requestOpts`. That may seems strange, but python-gnocchiclient uses the same header and acceptance test fails without it.

jtopjian:
@ozerovandrei I think the right delete function to look at is here:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L1159

There are a number of additional features available: https://gnocchi.xyz/rest.html#id16

The query schema for batch deletes is based on this: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L1304

I'm not sure how best to implement this. Given its complexity, I'm willing to either:

1. Defer this until later
2. Have a dedicated `BatchDelete` function which, again, can be implemented later.

ozerovandrei:
@jtopjian I guess it's better to implement a separate `BatchDelete` method so we will have a `Delete` for the simple deletion of a single resource without any request body and a complex `BatchDelete` with some `BatchDeleteOpts` to tune deletion parameters.

Also those methods will use different urls (`v1/resource/<type>` for a batch delete call and the `v1/resource/<type>/<uuid>`  for a simple delete).

jtopjian:
> Also those methods will use different urls (v1/resource/<type> for a batch delete call and the v1/resource/<type>/<uuid> for a simple delete).

Ah good point. I agree then, a separate `BatchDelete` function is best.

ozerovandrei:
@jtopjian this has been rebased from #17.

jtopjian:
@ozerovandrei The delete portion of this commit looks good. Once "update" has been merged, this is good to go. :)

ozerovandrei:
@jtopjian this has been rebased from the latest master.

jtopjian:
@ozerovandrei Looks good to me - thank you!

PR #19, opened 2018-01-18
Implement Gnocchi metric creation

ozerovandrei:
Add Gnocchi metric create request with tests and documentation.

`gnocchi metric create`

code in this PR:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L618
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/indexer/sqlalchemy.py#L670

### DIFF
--- a/acceptance/gnocchi/metric/v1/metrics.go
+++ b/acceptance/gnocchi/metric/v1/metrics.go
@@ -1,0 +1,28 @@
+package v1
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/utils/gnocchi/metric/v1/metrics"
+)
+
+// CreateMetric will create Gnocchi metric. An error will be returned if the
+// metric could not be created.
+func CreateMetric(t *testing.T, client *gophercloud.ServiceClient) (*metrics.Metric, error) {
+	// Metric will be created assuming that your Gnocchi's indexer installation was configured with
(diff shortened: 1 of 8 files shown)


ozerovandrei:
Note: right now acceptance test helper assumes that Gnocchi was installed with [default  archive  policies](https://gnocchi.xyz/operating.html#default-archive-policies).
Those policies usually created while configuring an indexer for Gnocchi with the command:
`gnocchi-upgrade` or more explicitly `gnocchi-upgrade --noskip-archive-policies-creation`.

I think that this if ok for now. But sometime in the future I can implement an archive policy create request and update this helper to also create some policy with random name.

jtopjian:
> I think that this if ok for now.

Yes, that sounds good.

Similar with #16, `created_by_project_id`, `created_by_user_id`, and `creator` don't seem to be valid fields. 

I'm getting an error when trying to create a metric using them. Are you seeing the same error? I noticed your unit fixtures have these fields, so perhaps these fields were in an older version of Gnocchi? If so, then I am OK leaving these fields in-place.

ozerovandrei:
@jtopjian yeah, sorry again for bad testing of all creation parameters.
I rechecked and deleted all invalid request options, unit tests and docs examples.

PRs for Gnocchi metrics are ready for a review.

ozerovandrei:
@jtopjian done.

Yeah, I think that we can build some complex stuff after merging all those changes into the one branch. I have no questions at this point.

jtopjian:
This looks good to me. Thank you for your patience with this :)

PR #20, opened 2018-01-18
Implement Gnocchi metric delete

ozerovandrei:
Add Gnocchi metric delete request with tests and documentation.

`gnocchi metric delete`

code in this PR:

https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L533
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/indexer/sqlalchemy.py#L1118


ozerovandrei:
@jtopjian those changes are rebased with master and can be reviewed.

jtopjian:
@ozerovandrei This looks good to me.

I tried looking into the `Accept` header to figure out why it's required. I think it has to do with Gnocci's use of Pecan. I've also confirmed that `python-gnocchiclient` is setting it on _every_ request by default, though it's possible some requests override the header.

PR #21, opened 2018-01-19
Implement Gnocchi measures list

ozerovandrei:
Add Gnocchi measures list request with tests and documentation.

`gnocchi measures show`

code in this PR:

API:
- https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L465

- Ceph: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/storage/ceph.py#L140
- Redis: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/storage/redis.py#L103
- File: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/storage/file.py#L142
- S3: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/storage/s3.py#L161
- Swift: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/storage/swift.py#L145

### DIFF
--- a/gnocchi/metric/v1/measures/requests.go
+++ b/gnocchi/metric/v1/measures/requests.go
@@ -1,0 +1,82 @@
+package measures
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+	"github.com/gophercloud/utils/gnocchi"
+)
+
+// ListOptsBuilder allows extensions to add additional parameters to the
+// List request.
+type ListOptsBuilder interface {
+	ToMeasureListQuery() (string, error)
+}
+
+// ListOpts allows to provide additional options to the Gnocchi measures List request.
+type ListOpts struct {
+	// Refresh can be used to force any unprocessed measures to be handled in the Gnocchi
+	// to ensure that List request returns all aggregates.
+	Refresh bool `q:"refresh"`
+
+	// Start is a start of time time range for the measures.
+	Start *time.Time
+
+	// Stop is a stop of time time range for the measures.
+	Stop *time.Time
+
+	// Aggregation is a needed aggregation method for returned measures.
+	// Gnocchi returns "mean" by default.
+	Aggregation string `q:"aggregation"`
+
+	// Granularity is a needed time between two series of measures to retreive.
+	// Gnocchi will response with all granularities for available measures by default.
+	Granularity string `q:"granularity"`
+
+	// Resample allows to select different granularity instead of those that were defined in the
+	// archive policy.
+	Resample string `q:"resample"`
+}
+
+// ToMeasureListQuery formats a ListOpts into a query string.
+func (opts ListOpts) ToMeasureListQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+
+	// delete
+	fmt.Printf("ToMeasureListQuery start: %s\n", q.String())
+
+	if opts.Start != nil {
+		start := strings.Join([]string{"start=", opts.Start.Format(gnocchi.RFC3339NanoNoTimezone)}, "")
+		q.RawQuery = strings.Join([]string{q.RawQuery, start}, "&")
+	}
+
+	if opts.Stop != nil {
+		stop := strings.Join([]string{"stop=", opts.Stop.Format(gnocchi.RFC3339NanoNoTimezone)}, "")
+		q.RawQuery = strings.Join([]string{q.RawQuery, stop}, "&")
+	}
+
+	// delete
+	fmt.Printf("ToMeasureListQuery finish: %s\n", q.String())
+
+	return q.String(), err
+}
+
+// List returns a Pager which allows you to iterate over a collection of
+// measures.
+// It accepts a ListOpts struct, which allows you to provide options to a Gnocchi measures List request.
+func List(c *gophercloud.ServiceClient, measureID string, opts ListOptsBuilder) pagination.Pager {
+	url := listURL(c, measureID)
+	if opts != nil {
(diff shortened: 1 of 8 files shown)


ozerovandrei:
Note: this can be reviewed, but acceptance test can be provided only after merging #19 because we need to create some metric to list it's measures.

ozerovandrei:
@jtopjian those changes are bebased from the current master and ready for a review.

I think that it's better no to implement separate acceptance test for that. We need a metric to list it's measures. And we need to create those measures.
So I want to implement `TestMeasuresCRUD` acceptance test, that will create metric, push some measures to it, list those measures and then delete this metric.
Metric creation were merged, metric deletion is waiting for a review in #20. Measures creation is waiting for a review in #22.

gnocchi/metric/v1/measures/requests.go
```diff
@@ -0,0 +1,75 @@
...
+	// Gnocchi will response with all granularities for available measures by default.
+	Granularity string `q:"granularity"`
+
+	// Resample allows to select different granularity instead of those that were defined in the
+	// archive policy.
+	Resample string `q:"resample"`
+}
+
+// ToMeasureListQuery formats a ListOpts into a query string.
+func (opts ListOpts) ToMeasureListQuery() (string, error) {
```
jtopjian: This function works, but I think we can clean it up a little. I only know of this because I was looking at the `net/url` package the other day :smile: 

```go
// ToMeasureListQuery formats a ListOpts into a query string.            
func (opts ListOpts) ToMeasureListQuery() (string, error) {              
  q, err := gophercloud.BuildQueryString(opts)                           
  params := q.Query()                                                    
                                                                         
  if opts.Start != nil {                                                 
    params.Add("start", opts.Start.Format(gnocchi.RFC3339NanoNoTimezone))
  }                                                                      
                                                                         
  if opts.Stop != nil {                                                  
    params.Add("stop", opts.Stop.Format(gnocchi.RFC3339NanoNoTimezone))  
  }                                                                      
                                                                         
  q = &url.URL{RawQuery: params.Encode()}                                
  return q.String(), err                                                 
}                                                                        
```

gnocchi/metric/v1/measures/results.go
```diff
@@ -0,0 +1,84 @@
...
+    [
+        "2017-01-08T10:05:00+00:00",
+        300.0,
+        58.0
+    ]
+]
+
+Helper unmarshals every nested array into the Measure type.
+*/
+func (r *Measure) UnmarshalJSON(b []byte) error {
```
jtopjian: This looks correct to me and this is what I would have done. Let's see if @jrperritt knows of an alternative way of doing this.

jrperritt: > This looks correct to me

I agree.

A few suggestions:
1. Check the length of the slice is 3 before trying to process it, otherwise it could `panic` (and if it isn't 3, we aren't prepared to process it anyway)
2. `TimeStamp` should be a `time.Time`; you can define a format to parse. See here: https://tip.golang.org/pkg/time/#Parse

gnocchi/metric/v1/measures/results.go
```diff
@@ -0,0 +1,84 @@
...
+	var s struct {
+		tmp
+	}
+	err := json.Unmarshal(b, &measuresSlice)
+	if err != nil {
+		return err
+	}
+
+	*r = Measure(s.tmp)
+	r.TimeStamp = measuresSlice[0].(string)
```
jrperritt: Additonally, each of the needs to check the assumed data type:
```go
var ok bool
r.TimeStamp, ok = measuresSlice[0].(string); !ok {
  // return an error because it's not a string
}
```

Without those checks, the code can `panic`. Remember, OpenStack is returning data that could be anything. Just because it's `[string, float64, float64]` now doesn't mean it won't change in the future, and if it does, we don't want programs to crash because of it.

ozerovandrei:
@jtopjian @jrperritt I've done cleanup of the `ToMeasureListQuery` method and added some checks to the `UnmarshalJSON` method.

gnocchi/metric/v1/measures/results.go
```diff
@@ -0,0 +1,115 @@
...
+	var measuresSlice []interface{}
+	err := json.Unmarshal(b, &measuresSlice)
+	if err != nil {
+		return err
+	}
+
+	// We need to check that a measure contains all needed data.
+	if len(measuresSlice) != 3 {
+		errMsg := fmt.Sprintf("got an invalid measure: %v", measuresSlice)
+		return errors.New(errMsg)
```
jtopjian: small nit: `fmt.Errorf` is more common than `errors.New`.

ozerovandrei:
@jtopjian I've fixed that.

jtopjian:
This looks good to me. @jrperritt can you review when you have a minute?

gnocchi/metric/v1/measures/results.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/gophercloud/utils/gnocchi"
+
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// Measure is an datapoint thats is composed with a timestamp and a value.
+type Measure struct {
+	// TimeStamp represents a timestamp of when measure was pushed into the Gnocchi.
+	TimeStamp time.Time `json:"-"`
```
jtopjian: This should be `Timestamp`.

jtopjian:
@ozerovandrei Just to confirm: this is good to go? It looks good to me, just want to make sure you didn't have any other commits planned :)

ozerovandrei:
@jtopjian yeah, this is totally ready 😃

PR #22, opened 2018-01-19
Implement Gnocchi measures create

ozerovandrei:
Add Gnocchi measures create request with tests and documentation.

That request implements pushing measures to a single metric, without batching.
Docs reference: [rest#push](https://gnocchi.xyz/rest.html#push)

`gnocchi measures add`

code in this PR:

API:
- https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L454

- Ceph: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/incoming/ceph.py#L193
- Redis: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/incoming/redis.py#L107
- File: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/incoming/file.py#L168
- S3: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/incoming/s3.py#L161
- Swift: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/incoming/swift.py#L100

### DIFF
--- a/gnocchi/metric/v1/measures/requests.go
+++ b/gnocchi/metric/v1/measures/requests.go
@@ -130,7 +130,7 @@ func Create(client *gophercloud.ServiceClient, metricID string, opts CreateOptsB
 		r.Err = err
 		return
 	}
-	_, r.Err = client.Post(pushURL(client, metricID), b["measures"], &r.Body, &gophercloud.RequestOpts{
+	_, r.Err = client.Post(createURL(client, metricID), b["measures"], &r.Body, &gophercloud.RequestOpts{
 		OkCodes: []int{202},
 		MoreHeaders: map[string]string{
 			"Accept": "application/json, */*",
--- a/gnocchi/metric/v1/measures/requests.go
+++ b/gnocchi/metric/v1/measures/requests.go
@@ -82,10 +82,10 @@ type CreateOptsBuilder interface {
 // MeasureOpts represents options of a single measure that can be created in the Gnocchi.
 type MeasureOpts struct {
 	// Timestamp represents a measure creation timestamp.
-	Timestamp time.Time
+	Timestamp time.Time `json:"-" required:"true"`
 
 	// Value represents a measure data value.
-	Value float64
+	Value float64 `json:"-" required:"true"`
 }
 
 // ToMap is a helper function to convert individual MeasureOpts structure into a sub-map.
--- a/gnocchi/metric/v1/measures/requests.go
+++ b/gnocchi/metric/v1/measures/requests.go
@@ -81,8 +81,8 @@ type CreateOptsBuilder interface {
 
 // MeasureOpts represents options of a single measure that can be created in the Gnocchi.
 type MeasureOpts struct {
-	// TimeStamp represents a measure creation timestamp.
-	TimeStamp time.Time
+	// Timestamp represents a measure creation timestamp.
+	Timestamp time.Time
 
 	// Value represents a measure data value.
 	Value float64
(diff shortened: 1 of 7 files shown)


ozerovandrei:
Note: this can be reviewed, but acceptance test can be provided only after merging #19  because we need to create some metric to push some measures to it.

ozerovandrei:
@jtopjian this is ready for a review. Although acceptance test contains commented line 
```
// defer DeleteMetric(t, client, metric.ID)
```
because that function lives in #20.

And this PR contains some tricky way to deal with converting timestamps of measures from the `time.Time` to the `string` type in the `ToMeasurePushMap`. Can I use that? Or can you please give an advice how can it be improved 😅

jtopjian:
@ozerovandrei That's what I would have done if I didn't know of any other method. The Gophercloud db/v1 database package does something very similar so we can ~steal~ leverage that:

https://github.com/gophercloud/gophercloud/blob/master/openstack/db/v1/databases/requests.go

Some other notes:

* I think we should rename `Push` to `Create`. I understand that the Gnocchi API uses the term "push", but it's still doing a `POST`. I feel like if we used the word "push", it might imply there's some sort of [http push](https://en.wikipedia.org/wiki/Push_technology) happening.

* I'm not a fan of the name `MeasureToPush`. Maybe `MeasureOpts`? I would have suggested calling it `CreateOpts` and then renaming the existing `CreateOpts` to `BatchCreateOpts` like in db/v1, but that would conflict with Gnocchi's "batch" support.

* `MeasureOpts` will have a method of `ToMap()` like in the above db/v1 link.

* `CreateOpts` will then simply be `[]MeasureOpts` and it will be implemented like `BatchCreateOpts` in the db/v1 link.

Thoughts?

ozerovandrei:
@jtopjian well that sounds like it will help to maintain some uniformity of a methods of the Gophercloud. So why not. All done.

ozerovandrei:
@jtopjian this is ready for a review now.

gnocchi/metric/v1/measures/requests.go
```diff
@@ -0,0 +1,140 @@
...
+	ToMeasureCreateMap() (map[string]interface{}, error)
+}
+
+// MeasureOpts represents options of a single measure that can be created in the Gnocchi.
+type MeasureOpts struct {
+	// TimeStamp represents a measure creation timestamp.
+	TimeStamp time.Time
+
+	// Value represents a measure data value.
+	Value float64
```
jtopjian: Value is required, right? If so, it should be:

```go
Value float64 `json:"value" required:"true"`
```

gnocchi/metric/v1/measures/requests.go
```diff
@@ -0,0 +1,140 @@
...
+
+// CreateOptsBuilder is needed to add measures to the Create request.
+type CreateOptsBuilder interface {
+	ToMeasureCreateMap() (map[string]interface{}, error)
+}
+
+// MeasureOpts represents options of a single measure that can be created in the Gnocchi.
+type MeasureOpts struct {
+	// TimeStamp represents a measure creation timestamp.
+	TimeStamp time.Time
```
jtopjian: It should be `Timestamp` instead of `TimeStamp`.

Also, is it required? If so, it should be:

```go
Timestamp time.Time `json:"-" required:"true"`
```

If it's not required, then just the `json` tag.

gnocchi/metric/v1/measures/requests.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// Value represents a measure data value.
+	Value float64
+}
+
+// ToMap is a helper function to convert individual MeasureOpts structure into a sub-map.
+func (opts MeasureOpts) ToMap() (map[string]interface{}, error) {
```
jtopjian: With the above changes made, you should be able to do something like:

```go
b, err := gophercloud.BuildRequestBody(opts, "")
if err != nil {
  return nil, err
}

if opts.Timestamp != nil {
  b["timestamp"] = opts.Timestamp.Format(gnocchi.RFC3339NanoNoTimezone),
}

return b
```

Or something like that - I haven't tested that out.

ozerovandrei: @jtopjian yeah we can do that.
Although we need to use the `*time.Time` for the `Timestamp` field in that case.

ozerovandrei:
@jtopjian I've rebased that PR with master and added the needed changes.

jtopjian:
@ozerovandrei This looks good to me. Nice work on this one 😄

PR #23, opened 2018-01-24
Implement Gnocchi measures batch create via metrics IDs

ozerovandrei:
Add ability to create measures of a different metrics via metrics IDs references.
Add unit and acceptance tests and a documentation example.

That request implements pushing measures to different metrics via batch request to metric IDs.
Those changes won't provide the ability that allow user to batch measures via resource IDs and metric names.

Docs reference: [rest.html#batch](https://gnocchi.xyz/rest.html#batch)

code in this PR:

API:

- https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L1614

- Ceph: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/incoming/ceph.py#L81
- Redis: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/incoming/redis.py#L52
- File: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/incoming/file.py#L168
- S3: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/incoming/s3.py#L161
- Swift: https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/incoming/swift.py#L100

### DIFF
--- a/gnocchi/metric/v1/measures/requests.go
+++ b/gnocchi/metric/v1/measures/requests.go
@@ -134,3 +134,58 @@ func Create(client *gophercloud.ServiceClient, metricID string, opts CreateOptsB
 	})
 	return
 }
+
+// CreateBatchMetricsOptsBuilder is needed to add measures to the CreateBatchMetrics request.
+type CreateBatchMetricsOptsBuilder interface {
+	ToMeasureCreateBatchMetricsMap() (map[string]interface{}, error)
+}
+
+// CreateBatchMetricsOpts specifies a parameters for creating measures for different metrics in a single request.
+type CreateBatchMetricsOpts struct {
+	// BatchOpts is a map of metric ids and corresponding measures that needs to be created.
+	BatchOpts map[string][]MeasureOpts
+}
+
+// ToMeasureCreateBatchMetricsMap constructs a request body from CreateBatchMetricsOpts.
+func (opts CreateBatchMetricsOpts) ToMeasureCreateBatchMetricsMap() (map[string]interface{}, error) {
+	// measures is an internal map representation of the MeasureOpts struct.
+	type measureOpts map[string]interface{}
+
+	// measures is a slice of measures maps.
+	var measures []measureOpts
+
+	// batchOpts is a internal representation of the BatchOpts, where MeasureOpts represented as maps.
+	batchOpts := make(map[string][]measureOpts)
+
+	// Populate batchOpts.
+	for k, v := range opts.BatchOpts {
+		measures = make([]measureOpts, len(v))
+		for i, m := range v {
+			measureMap, err := m.ToMap()
+			if err != nil {
+				return nil, err
+			}
+			measures[i] = measureMap
+		}
+		batchOpts[k] = measures
+	}
+
+	return map[string]interface{}{"batchMeasures": batchOpts}, nil
+}
+
+// CreateBatchMetrics requests the creation of a new measures for different metrics.
+func CreateBatchMetrics(client *gophercloud.ServiceClient, opts CreateBatchMetricsOptsBuilder) (r CreateBatchMetricsResult) {
+	b, err := opts.ToMeasureCreateBatchMetricsMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+
+	_, r.Err = client.Post(createBatchMetricsURL(client), b["batchMeasures"], &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{202},
+		MoreHeaders: map[string]string{
+			"Accept": "application/json, */*",
+		},
+	})
+	return
+}
--- a/gnocchi/metric/v1/measures/requests.go
+++ b/gnocchi/metric/v1/measures/requests.go
@@ -135,53 +135,69 @@ func Create(client *gophercloud.ServiceClient, metricID string, opts CreateOptsB
 	return
 }
 
-// BatchMetricsOptsBuilder is needed to add measures to the BatchMetrics request.
-type BatchMetricsOptsBuilder interface {
-	ToMeasureBatchMetricsMap() (map[string]interface{}, error)
+// BatchCreateMetricsOptsBuilder is needed to add measures to the BatchCreateMetrics request.
+type BatchCreateMetricsOptsBuilder interface {
+	ToBatchCreateMetricsMap() (map[string]interface{}, error)
 }
 
-// BatchMetricsOpts specifies a parameters for creating measures for different metrics in a single request.
-type BatchMetricsOpts struct {
-	// BatchOpts is a map of metric ids and corresponding measures that needs to be created.
-	BatchOpts map[string][]MeasureOpts
+// BatchCreateMetricsOpts specifies a parameters for creating measures for different metrics in a single request.
+type BatchCreateMetricsOpts struct {
+	BatchMetricsOpts []MetricOpts
 }
 
-// ToMeasureBatchMetricsMap constructs a request body from BatchMetricsOpts.
-func (opts BatchMetricsOpts) ToMeasureBatchMetricsMap() (map[string]interface{}, error) {
-	// measures is an internal map representation of the MeasureOpts struct.
-	type measureOpts map[string]interface{}
+// MetricOpts represents measures of a single metric of the BatchCreateMetrics request.
+type MetricOpts struct {
+	ID       string
+	Measures []MeasureOpts
+}
 
+// ToMap is a helper function to convert individual MetricOpts structure into a sub-map.
+func (opts MetricOpts) ToMap() (map[string]interface{}, error) {
 	// measures is a slice of measures maps.
-	var measures []measureOpts
+	measures := make([]map[string]interface{}, len(opts.Measures))
 
-	// batchOpts is a internal representation of the BatchOpts, where MeasureOpts represented as maps.
-	batchOpts := make(map[string][]measureOpts)
+	// metricOpts is an internal map representation of the MetricOpts struct.
+	metricOpts := make(map[string]interface{})
 
-	// Populate batchOpts.
-	for metricID, metricMeasures := range opts.BatchOpts {
-		measures = make([]measureOpts, len(metricMeasures))
-		for i, measure := range metricMeasures {
-			measureMap, err := measure.ToMap()
-			if err != nil {
-				return nil, err
-			}
-			measures[i] = measureMap
+	for i, measure := range opts.Measures {
+		measureMap, err := measure.ToMap()
+		if err != nil {
+			return nil, err
 		}
-		batchOpts[metricID] = measures
+		measures[i] = measureMap
+	}
+	metricOpts[opts.ID] = measures
+
+	return metricOpts, nil
+}
+
+// ToBatchCreateMetricsMap constructs a request body from BatchCreateMetricsOpts.
+func (opts BatchCreateMetricsOpts) ToBatchCreateMetricsMap() (map[string]interface{}, error) {
+	// batchCreateMetricsOpts is an internal representation of the BatchCreateMetricsOpts struct.
+	batchCreateMetricsOpts := make(map[string]interface{})
+
+	for _, metricOpts := range opts.BatchMetricsOpts {
+		metricOptsMap, err := metricOpts.ToMap()
+		if err != nil {
+			return nil, err
+		}
+		for k, v := range metricOptsMap {
+			batchCreateMetricsOpts[k] = v
+		}
 	}
 
-	return map[string]interface{}{"batchMeasures": batchOpts}, nil
+	return map[string]interface{}{"batchCreateMetrics": batchCreateMetricsOpts}, nil
 }
 
-// BatchMetrics requests the creation of a new measures for different metrics.
-func BatchMetrics(client *gophercloud.ServiceClient, opts BatchMetricsOptsBuilder) (r BatchMetricsResult) {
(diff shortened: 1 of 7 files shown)


ozerovandrei:
@jtopjian this is also ready for a review.

The request `CreateBatchMetrics` have such name because those changes only implements measures batching via metric IDs: [rest.html#batch](https://gnocchi.xyz/rest.html#batch).
I want to also implement batching via resource IDs and metric names (next example under the [rest.html#batch](https://gnocchi.xyz/rest.html#batch)) for the main issue [#698](https://github.com/gophercloud/gophercloud/issues/698) (that will be the last PR for that issue). And that request will have the name `CreateBatchResourcesMetrics`.

I'm not really sure right now how I feel about those names 😄 

Also there is some tricky method to look at: `ToMeasureCreateBatchMetricsMap`.

jtopjian:
@ozerovandrei Thank you for the amount of detail used in this one (docs, tests). 

This looks correct to me, but I think we might want to rename some of the functions or possibly move this to the `metrics` package or maybe even a dedicated `batch` package. When the functions have words like measure, batch, create, and metrics all in one name, it might be a sign to try to simplify it somehow. I'm not sure yet. Any thoughts?

I'm currently attending some conferences in Europe so I only have a little time to work on PRs. I'll think about this over the next day or two.

ozerovandrei:
@jtopjian yeah I think that those names are too huge 😃 

In the last PRs #23 #24 there are two big main requests: `measures.CreateBatchMetrics` and `measures.CreateBatchResourcesMetrics`.
We can still use the `measures` package but rename those requests to `measures.BatchMetrics` and `measures.BatchResources`. So we still have `measures` word to emphasize that we are batching measures. I think that measure is the thing that really matters in those requests.

In case of renaming that package to the `batch` we can use something like `batch.MeasuresToMetrics` and `batch.MeasuresToResources`.

But more important thing I guess is the review on the whole. So all the logic inside `requests.go` helpers like `ToMeasureCreateBatchMetricsMap` and `ToMeasureCreateBatchResourcesMetricsMap` is good to go and just needs some renaming.
And also don't forget to look at new helpers in the `acceptance/gnocchi/metric/v1/measures.go`, they also do some work with maps and stuff 😃

jtopjian:
> But more important thing I guess is the review on the whole. So all the logic inside requests.go helpers like ToMeasureCreateBatchMetricsMap and ToMeasureCreateBatchResourcesMetricsMap is good to go and just needs some renaming.

Indeed, yes. I did a cursory review and everything looks good. In addition...

> And also don't forget to look at new helpers in the acceptance/gnocchi/metric/v1/measures.go, they also do some work with maps and stuff 

Yeah, really nice work on that, too. There's a lot of good work here so I apologize that focusing on the naming/organizing overshadowed that.

ozerovandrei:
@jtopjian for now I've renamed those requests to `measures.BatchMetrics` and `measures.BatchResources`. 
I think that it's much better now. I don't know how we can improve that. Again, it's about measures so `measures` package is a good place for those requests.

jtopjian:
@ozerovandrei I had some time to review everything. Here are my thoughts on the names:

* We can't use "batch" on its own since it's possible to batch-delete resources. If batching was only for creating, we could get away with using only the word "batch". 
* This is probably an English nuance, but it's better to use `BatchCreate` rather than `CreateBatch`. 
* Since it's possible to do batch creation of measures with _both_ metrics and resources, we can't simply have `BatchCreate`.

So I propose the following:

* `measures.BatchCreateMetrics` (`BatchCreateMetricsOptsBuilder`, `ToMeasureBatchCreateMetricsMap`, `BatchCreateMetricsOpts`, etc)
* `measures.BatchCreateResources` (or perhaps we should do the fully explicit `BatchCreateResourceMetrics` just like how the API URL is...)

I realize we're back to the original names, but with "Batch" and "Create" swapped :) But after reviewing everything, I think you were originally correct. The names are long but it matches the Gnocchi API and prevents confusion on whether it is a batch _create_ or batch _delete_. I apologize for the extra work -- I just needed some time to look at this in more detail.

One thing I haven't had time to think about is the `BatchOps` field in the `BatchMetricsOpts` (`BatchCreateMetricsOps`) struct. I wonder if there's the ability to have something more type-specific instead of `map[string][]MeasureOpts`. Maybe something like `BatchMetrics` which has an ID field and a `[]MeasureOpts` field. I haven't put a lot of thought into this, though, so I'm not sure.

I apologize for going around in a circle. I didn't have a lot of time to think about the names and was more thinking out loud with my last comment. I'm sorry for the extra work.

But again, there's a lot of good work in this PR and I do appreciate the time and thought you've put into it :)

Let me know if you have any questions and/or want to discuss more.

ozerovandrei:
@jtopjian yeah those names are more similar to Gnocchi's internal API controllers. And maybe Gnocchi will have more stuff for batching (like batch creating new empty resources or something like that), so we will need to do some renaming in the future. Which is bad.

About `measures.BatchCreateResources` (#24). I think that `measures.BatchCreateResourceMetrics` or even `measures.BatchCreateResourcesMetrics` is better because we don't create resources in this request. Gnocchi can create metrics for those resources with the request option `create_metrics=True` and will create measures, but resources should be here already.
So `BatchCreateResources` sounds more like we're trying to create resources and that's not what this requests is really doing.

About `BatchOps`. Now it's a map with a `string` key that is representing an `ID` of a metric in the `measures.BatchCreateMetrics` request:
```
{
  "99f7eb88-d71b-4308-8116-f8afa5b87ee0": [
    {
      "timestamp": "2014-10-06T16:12:12",
      "value": 3
    },
    {
      "timestamp": "2014-10-06T18:14:52",
      "value": 4
    }
  ],
  "e8e7bde8-849b-4f96-874e-fc21589e6cca": [
    {
      "timestamp": "2014-10-06T14:34:12",
      "value": 12
    },
    {
      "timestamp": "2014-10-06T14:34:20",
      "value": 2
    }
  ]
}
```

and a `Name` of a metric in the `measures.BatchCreateResourcesMetrics` request:
```
{
  "15e9c872-7ca9-11e4-a2da-2fb4032dfc09": {
    "cpu.util": [
      {
        "timestamp": "2014-10-06T14:34:12",
        "value": 12
      },
      {
        "timestamp": "2014-10-06T14:34:20",
        "value": 2
      }
    ]
  },
  "6f24edd9-5a2f-4592-b708-ffbed821c5d2": {
    "cpu.util": [
      {
        "timestamp": "2014-10-06T14:34:12",
        "value": 6
      },
      {
        "timestamp": "2014-10-06T14:34:20",
        "value": 25
      }
    ]
  }
}
```

So, it's re-used in both requests and if we will implement it as a struct we will need an `ID` and a `Name` (and `[]MeasureOpts`, of course) in that struct. That looks not so elegant.

Also when I look at that JSON objects with measures, all I can think of is the mapping between a metric name/id and a list of measures. And because of that I'm seeing a `map[string][]MeasureOpts` on which we can iterate and do other hash-table-specific tricks 😃 

Is there any benefits of using a struct instead of a map in that case?

ozerovandrei:
@jtopjian oh well. I wrote all that stuff and then I went to the `master` branch of the Gnocchi. So, they have implemented an ability to provide the `archive_policy_name` in the second batch type: [gnocchi/pull/655](https://github.com/gnocchixyz/gnocchi/pull/655)
[doc example](https://gnocchi.xyz/master/rest.html#push).

So we will really need different structs in that case because the second batch type transforms to this (`archive_policy_name` is optional):
```
{
  "15e9c872-7ca9-11e4-a2da-2fb4032dfc09": {
    "cpu.util": {
      "archive_policy_name": "short",
      "measures": [
        {
          "timestamp": "2014-10-06T14:34:12",
          "value": 12
        },
        {
          "timestamp": "2014-10-06T14:34:20",
          "value": 2
        }
      ]
    }
  },
  "6f24edd9-5a2f-4592-b708-ffbed821c5d2": {
    "cpu.util": {
      "archive_policy_name": "short",
      "measures": [
        {
          "timestamp": "2014-10-06T14:34:12",
          "value": 6
        },
        {
          "timestamp": "2014-10-06T14:34:20",
          "value": 25
        }
      ]
    }
  }
}
```

Well. It looks like it will break the current implementation of the `measures.BatchCreateResourcesMetrics` request 😞 
What we should do in that case?

ozerovandrei:
_Summary_
I want to use `measures.BatchCreateMetrics` and `measures.BatchCreateResourcesMetrics` for naming and stay with a map for `BatchOps`. Also I think it's better to implement the current stable version of the #24 batch request and ignore the `master` release of Gnocchi for now.

jtopjian:
> Sorry for a long text.

Absolutely no problem at all. 

> Is there any benefits of using a struct instead of a map in that case?

Yes, the benefit of using a type-defined struct is twofold:

1. You're modeling the data in detail which provides the user/dev with a better understanding of the fields. 
2. You are creating type restrictions which enforces only certain data to be included in fields. This is one of the biggest benefit of using Go.

As a dramatic/exaggerated example: consider if Gophercloud used only `map[string]interface{}` for _everything_: opts and results. Admittedly this is a tempting thing to do some days 😄 but Gophercloud would lose a lot of its stability.

Of course there are times when it's appropriate to use a generic map. 

> Also I think it's better to implement the current stable version of the #24 batch request and ignore the master release of Gnocchi for now.

One issue with this is that it would be a big change later on. I would much prefer to use a struct in preparation, even if the fields available in master aren't used here. This way, if someone wanted to add the fields later on, they would only have to add fields to a struct rather than refactor a generic map.

With this in mind, I'd like to split discussion of #23 and #24 for simplicity. So only focusing on `BatchCreateMetrics`:

1. Technically you can do this:

```go
type BatchCreateMetricsOpts map[string][]MeasureOpts
```

This is very similar to what is done in the `db/v1/databases` package. This method is a good choice if we're sure we'll never have to add fields `BatchCreateMetricsOpts`.

2. Leave `BatchCreateMetricsOpts` as-is. It only has a single field (`BatchOpts` - which I think we need to rename to _something_ but I'm not sure what yet) but it looks similar to other `CreateOpts` structs.

3. Try something like:

```go
type BatchCreateMetricsOpts struct {
  MetricsOpts []MetricsOpts
}

type MetricsOpts struct {
  ID string
  MeasureOpts []MeasureOpts
}
```

This method is very verbose which might be good or bad.

I'm a little busy with some other work and it's taken me all day to type up this comment. I haven't had time to think about which method would be best. You're more than welcome to test the methods out and see which one you like, though it would probably be best not to commit anything until we've agreed (unless you don't mind the extra work :)

ozerovandrei:
@jtopjian thank you for the explanation about the types.

Well.
The first variant is bad because we can never be sure that we won't need some more fields in the future.
With the second variant we _maybe_ can try to re-use it inside the #24, because we will need to implement this:
```
"measures": [
  {
    "timestamp": "2014-10-06T14:34:12",
    "value": 12
  },
  {
    "timestamp": "2014-10-06T14:34:20",
    "value": 2
  }
]
```
So in #23 we will already have that structure in `BatchOpts map[string][]MeasureOpts`.
We can't re-use the third variant for #24 because it won't have the id of a metric.

ozerovandrei:
@jtopjian oh no, that's a mistake. We can't use a map for that field in the #24. It will be just a field in the struct with a []MeasureOpts as a value.

So we need a second or a third variant. Third is better if metrics batch will have additional fields in the future. So we can add them easily inside this:
```
type MetricsOpts struct {
  ID string
  MeasureOpts []MeasureOpts
}
```

ozerovandrei:
@jtopjian I've played with the third variant of the implementation. You can check how it looks.

gnocchi/metric/v1/measures/requests.go
```diff
@@ -134,3 +134,74 @@ func Create(client *gophercloud.ServiceClient, metricID string, opts CreateOptsB
 	})
 	return
 }
+
+// BatchCreateMetricsOptsBuilder is needed to add measures to the BatchCreateMetrics request.
+type BatchCreateMetricsOptsBuilder interface {
+	ToBatchCreateMetricsMap() (map[string]interface{}, error)
```
jtopjian: This should be `ToMeasuresBatchCreateMetricsMap`. The format of the interface name is `To-ResourcePkg-Action-RequestType`.

ozerovandrei:
@jtopjian I've fixed the name.
I've checked the `type BatchCreateMetricsOpts []MetricOpts` and i think it's providing at least on benefit - options for the `BatchCreateMetrics` are little cleaner now:
```
createOpts := measures.BatchCreateMetricsOpts{
  {
    ID: "777a01d6-4694-49cb-b86a-5ba9fd4e609e",
    Measures: []measures.MeasureOpts{
      {
        Timestamp: &currentTimestamp,
        Value:     200,
      },
      {
        Timestamp: &pastHourTimestamp,
        Value:     300,
      },
    },
  },
  {
    ID: "6dbc97c5-bfdf-47a2-b184-02e7fa348d21",
    Measures: []measures.MeasureOpts{
      {
        Timestamp: &currentTimestamp,
        Value:     111,
      },
      {
        Timestamp: &pastHourTimestamp,
        Value:     222,
      },
    },
  },
}
```

I've commited those changes.

jtopjian:
@ozerovandrei This looks really good and I really like the final result.

Is this good to merge or is there anything else you wanted to do?

ozerovandrei:
@jtopjian let's merge it.

jtopjian:
Sounds good. Nice work on this :)

PR #24, opened 2018-01-26
Implement Gnocchi measures batch create via resource IDs and metric names

ozerovandrei:
Add ability to create measures of a different metrics via resource IDs and metrics names references.
Add unit and acceptance tests and a documentation example.

That request implements pushing measures to different metrics via batch request to resource IDs and metric names.

Docs reference: [rest.html#batch (second example)](https://gnocchi.xyz/stable_4.2/rest.html#batch)

code in this PR:

API:

- https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/rest/api.py#L1567

- Ceph: https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/incoming/ceph.py#L81
- Redis: https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/incoming/redis.py#L52
- File: https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/incoming/file.py#L166
- S3: https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/incoming/s3.py#L161
- Swift: https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/incoming/swift.py#L100

### DIFF
--- a/acceptance/gnocchi/metric/v1/measures.go
+++ b/acceptance/gnocchi/metric/v1/measures.go
@@ -88,8 +88,6 @@ func CreateBatchResourcesMetricsMeasures(t *testing.T, client *gophercloud.Servi
 	batchOpts := make(map[string]map[string][]measures.MeasureOpts)
 
 	for resourceID, metricNames := range batchResourcesMetrics {
-		// DELETE
-		t.Logf("CreateBatchResourcesMetricsMeasures resourceID: %v\n", resourceID)
 		metricMap := make(map[string][]measures.MeasureOpts)
 		for _, metricName := range metricNames {
 			metricMap[metricName] = []measures.MeasureOpts{
--- a/acceptance/gnocchi/metric/v1/measures.go
+++ b/acceptance/gnocchi/metric/v1/measures.go
@@ -31,8 +31,7 @@ func CreateMeasures(t *testing.T, client *gophercloud.ServiceClient, metricID st
 	}
 
 	t.Logf("Attempting to create measures inside a Gnocchi metric %s", metricID)
-
-	if err := measures.Create(client, metricID, createOpts).ExtractErr(); err != nil && err.Error() != "EOF" {
+	if err := measures.Create(client, metricID, createOpts).ExtractErr(); err != nil {
 		return err
 	}
 
@@ -67,8 +66,7 @@ func MeasuresBatchCreateMetrics(t *testing.T, client *gophercloud.ServiceClient,
 	}
 
 	t.Logf("Attempting to create measures inside Gnocchi metrics via batch request")
-
-	if err := measures.BatchCreateMetrics(client, createOpts).ExtractErr(); err != nil && err.Error() != "EOF" {
+	if err := measures.BatchCreateMetrics(client, createOpts).ExtractErr(); err != nil {
 		return err
 	}
 
@@ -125,7 +123,7 @@ func MeasuresBatchCreateResourcesMetrics(t *testing.T, client *gophercloud.Servi
 	}
 
 	t.Logf("Attempting to create measures inside Gnocchi metrics via batch request with resource IDs")
-	if err := measures.BatchCreateResourcesMetrics(client, createOpts).ExtractErr(); err != nil && err.Error() != "EOF" {
+	if err := measures.BatchCreateResourcesMetrics(client, createOpts).ExtractErr(); err != nil {
 		return err
 	}
 
--- a/acceptance/gnocchi/metric/v1/measures.go
+++ b/acceptance/gnocchi/metric/v1/measures.go
@@ -76,10 +76,10 @@ func MeasuresBatchCreateMetrics(t *testing.T, client *gophercloud.ServiceClient,
 	return nil
 }
 
-// CreateBatchResourcesMetricsMeasures will create measures inside different metrics via batch request to resource IDs.
+// BatchResourcesMeasures will create measures inside different metrics via batch request to resource IDs.
 // The batchResourcesMetrics arguments is a mapping between resource IDs and corresponding metric names.
 // An error will be returned if measures could not be created.
-func CreateBatchResourcesMetricsMeasures(t *testing.T, client *gophercloud.ServiceClient, batchResourcesMetrics map[string][]string) error {
+func BatchResourcesMeasures(t *testing.T, client *gophercloud.ServiceClient, batchResourcesMetrics map[string][]string) error {
 	currentTimestamp := time.Now().UTC()
 	pastHourTimestamp := currentTimestamp.Add(-1 * time.Hour)
 	currentValue := float64(tools.RandomInt(100, 200))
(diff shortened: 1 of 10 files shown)


ozerovandrei:
Note: this can be reviewed, but the acceptance test "TestBatchResourcesMetricsMeasuresCreation" won't cleanup resources for now, because it needs the #18 changes.

ozerovandrei:
@jtopjian this has been rebased from #23.

jtopjian:
@ozerovandrei Per comments in #23, with the discovery that `BatchCreateResourceMetricsOpts` _will_ have additional fields, it's probably best to design the opts struct with defined fields. In addition to the `archive_policy_name`, there's also `unit`: https://github.com/gnocchixyz/gnocchi/blob/081e0513e68fb39630f2abf22fac0e617bce6af6/gnocchi/tests/functional/gabbits/batch-measures.yaml#L183-L209

I understand what you mean that these fields are still in master and haven't been officially released, but we should prepare for their addition. It's even possible that they will be published while we're working on this :)

jtopjian:
@ozerovandrei Yeah, I think the above is the right solution here. Similar to my comment in #23, is there any benefit of doing:

```go
type BatchCreateResourcesMetricsOpts []BatchResourcesMetricsOpts
```

? 

Again, just thinking out loud.

ozerovandrei:
@jtopjian same as in #23 I guess that options for the `BatchCreateResourcesMetrics()` will be more readable if we have the 
```
type BatchCreateResourcesMetricsOpts []BatchResourcesMetricsOpts
```

ozerovandrei:
@jtopjian this is ready for a review.

1. I've done a mistake, we can't use the
```
type BatchCreateResourcesMetricsOpts []BatchResourcesMetricsOpts
```
because `BatchCreateResourcesMetricsOpts` also have an additional field `CreateMetrics` that is used for adding a query option `create_metrics=True`.
So my current variant is:
```
type BatchCreateResourcesMetricsOpts struct {
  CreateMetrics bool `q:"create_metrics"`
  BatchResourcesMetrics []BatchResourcesMetricsOpts
}
```
```
type BatchResourcesMetricsOpts struct {
  ResourceID string
  ResourcesMetrics []ResourcesMetricsOpts
}
```
```
type ResourcesMetricsOpts struct {
  MetricName string `json:"-" required:"true"`
  ArchivePolicyName string
  Unit string
  Measures []MeasureOpts `json:"-" required:"true"`
}
```

Both a `BatchResourcesMetricsOpts` and a `ResourcesMetricsOpts` have the `ToMap()` helpers methods. 

2. I've added `json:"-" required:"true"` tag to some field of `ResourcesMetricsOpts` and `MetricOpts`(#23). As far as I remember we need `required` tag in most of the cases. And we still need them in #23 and #24 options, right?

3. Acceptance test and it's helpers were updated too. But to run it make sure that you've installed the Gnocchi from the `master` branch. Because the new batching format were merged only in `master` and it's not in any release yet.

jtopjian:
@ozerovandrei 

1. No problem - makes sense.

2. Technically yes. But in order for the `required` tag to work, you have to call [`gophercloud.BuildRequestBody`](https://github.com/gophercloud/gophercloud/blob/master/params.go#L35). If calling that function complicates things, then you can have a manual check to make sure the field is set.

3. That's fine. Question: is `measures.BatchCreateResourcesMetrics` only compatible with the current master branch? Or will it still work with the current stable release as long as the new fields aren't set?

4. Fair point. I'll add it to my TODO.

ozerovandrei:
@jtopjian adding [`gophercloud.BuildRequestBody`](https://github.com/gophercloud/gophercloud/blob/master/params.go#L35) will be pain. So I've added manual checks into `ToMap()` methods.

About compatibility. Gnocchi `master` can accept both formats. But `measures.BatchCreateResourcesMetrics` can _only_ work with the `master` version because request body structure is different even when you don't set `archive_policy_name` and `unit` fields.

In `master` measures reside inside the `measures` field:
```
{
  "15e9c872-7ca9-11e4-a2da-2fb4032dfc09": {
    "cpu.util": {
      "measures": [
        {
          "timestamp": "2014-10-06T14:34:12",
          "value": 12
        },
        {
          "timestamp": "2014-10-06T14:34:20",
          "value": 2
        }
      ]
    }
 }
```

And in `stable/4.1` they reside inside metric name:
```
{
  "15e9c872-7ca9-11e4-a2da-2fb4032dfc09": {
    "cpu.util": [
      {
        "timestamp": "2014-10-06T14:34:12",
        "value": 12
      },
      {
        "timestamp": "2014-10-06T14:34:20",
        "value": 2
      }
    ]
  }
}
```

jtopjian:
@ozerovandrei I just noticed the `EOF` checks being done in the unit tests... it looks like you aren't writing any fixtures to the mock http server.

Fixtures are missing for 

* `TestCreateMeasures`
* `TestBatchCreateMetrics`
* `TestBatchCreateResourcesMetrics`

Once you have these fixtures in place, you can remove the `EOF` checks.

ozerovandrei:
@jtopjian that's because real Gnocchi returns empty response for those requests:
[rest.html#push](https://gnocchi.xyz/rest.html#push), [rest.html#batch](https://gnocchi.xyz/rest.html#batch).

I also added checks for EOF in documentation and acceptance tests for that requests like that:
```
if err := measures.BatchCreateResourcesMetrics(gnocchiClient, createOpts).ExtractErr(); err != nil && err.Error() != "EOF" {
  panic(err)
}
```

jtopjian:
Checking for `EOF` should not be required. If you're seeing `EOF` _errors_ then something is wrong somewhere.

It's fine that Gnocchi returns an empty response. The problem is that you aren't sending a _request_ fixture. For example:

* https://github.com/gophercloud/utils/blob/master/gnocchi/metric/v1/resources/testing/fixtures.go#L126-L133
* https://github.com/gophercloud/utils/blob/master/gnocchi/metric/v1/resources/testing/requests_test.go#L103

jtopjian:
Thinking this over, if there is no response being sent, then you might want to approach the unit tests differently. The goal of these unit tests would be to ensure the create options are converted into the proper JSON request bodies. There might be precedence for this somewhere in Gophercloud.

jtopjian:
> Gophercloud acceptance test EOF check.

This is to cover up a bug that happens maybe 1/1000 times of running the test. Note that this is in the acceptance test and not the unit test.

The problem with this PR's unit test is the reverse -- we're covering up an EOF but the EOF does not happen in the acceptance tests :)

ozerovandrei:
@jtopjian I've removed EOF and added valid fixtures with `TestJSONRequest` method.

jtopjian:
@ozerovandrei Awesome - thank you 😄 

This looks good to me -- very nice work, especially with the handling of the query parameter + request body fields. Great acceptance test, too. 

It's kind of funny that 4.2 was released while you were working on this, too 😄  The way I see it, the `interface`'s will allow someone to rewrite the create opts to accept the old format if they need to. But that's a limited-time problem. The new format will be longer-living and makes more sense to include it here.

Issue #25, opened 2018-03-26
Support token authentication from clouds.yaml


leesdolphin:
I'm using Terraform to deploy my instances and I want to store my authentication token in my `clouds.yaml` so I don't have to keep authenticating with my password. 

The `clouds.yaml` looks like this and works with the openstack CLI:

```yaml
clouds:
  example:
    auth_type: token
    auth:
      auth_url: https://auth.cloud.example.com:5000/v3
      token: my-secret-token
      project_id: 1234
      project_name: "example.net"
    region_name: "region"
    interface: "public"
    identity_api_version: 3
```

I've talked with some of our cloud's team about making documentation improvements to openstack itself.

jtopjian:
@leesdolphin Thank you for reporting this. It looks like token auth is indeed possible with `clouds.yaml`:

https://github.com/openstack/os-client-config/blob/480bfc27224a675a483a3c00a2c831054a1ffdc6/os_client_config/tests/test_config.py#L755-L768

I'll look at getting this added.

jtopjian:
@leesdolphin This has been implemented in #27. It'll be supported in Terraform shortly. I apologize for the delay.

PR #26, opened 2018-03-27
Adding token authentication to clouds.yaml

jtopjian:
For #25 

I removed the requirement to have a username and password set since token-based authentication complicates that. Invalid combinations will now be handled by Gophercloud and any errors will propagate up.

There are other possible `auth_types` such as `v3password` and `v2password`. Support for those can be added as needed. For now, `clouds.yaml` files which specify them will implicitly work if all other settings (username, password, and auth_url) are set correctly.

I've tested this locally with Terraform and can confirm the `clouds.yaml` format given in #25 works.

/cc @dklyle

### DIFF
--- a/openstack/clientconfig/requests.go
+++ b/openstack/clientconfig/requests.go
@@ -168,19 +168,23 @@ func AuthOptions(opts *ClientOpts) (*gophercloud.AuthOptions, error) {
 		ao.DomainName = v
 	}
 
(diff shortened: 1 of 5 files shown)


jtopjian:
@jrperritt ping on this one when you have a moment

jrperritt:
closed in favor of #27

PR #27, opened 2018-04-06
clientconfig refactor

jtopjian:
This commit is a large refactor of the clientconfig package.

It attempts to become a single point of entry for all OpenStack
authentication, supporting authentication from clouds.yaml,
explicit authentication parameters, and environment variables.

Supersedes #26 

Requires https://github.com/gophercloud/gophercloud/pull/896

For https://github.com/gophercloud/gophercloud/issues/252

### DIFF
--- a/openstack/clientconfig/requests.go
+++ b/openstack/clientconfig/requests.go
@@ -3,6 +3,7 @@ package clientconfig
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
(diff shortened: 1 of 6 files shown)


jtopjian:
@jrperritt ping for review at your convenience

openstack/clientconfig/requests.go
```diff
@@ -18,6 +19,14 @@ type ClientOpts struct {
 
 	// EnvPrefix allows a custom environment variable prefix to be used.
 	EnvPrefix string
+
+	// AuthType specifies the type of authentication to use.
+	// By default, this is "password".
+	AuthType string
```
jrperritt: Can we make this a more specific type than `string`?

openstack/clientconfig/requests.go
```diff
@@ -18,6 +19,14 @@ type ClientOpts struct {
...
+
+	// Auth defines the authentication information needed to
+	// authenticate to a cloud when clouds.yaml isn't used.
+	Auth *Auth
```
jrperritt: How about `AuthInfo` for this (both these, I guess)?

openstack/clientconfig/requests.go
jtopjian: Yes, good catch.

openstack/clientconfig/requests.go
jtopjian: Yes, makes sense.

jtopjian:
@jrperritt Thank you for the review. Comments addressed and ready for another round.

jrperritt:
Looks good. Good job on this

Issue #28, opened 2018-04-10
Gnocchi measures can contain zero value


ozerovandrei:
Currently `measures` package doesn't allow us to send zero value in the Gnocchi measure.
But it's a completely normal situation to have zero values in measure for some periods of time.

Gnocchi API doesn't deny zero values:
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L454
https://github.com/gnocchixyz/gnocchi/blob/stable/4.1/gnocchi/rest/api.py#L426

jtopjian:
Fixed in #29

PR #29, opened 2018-04-10
Do not set required tag on the MeasureOpts.Value

ozerovandrei:
Gnocchi measures value can contain 0. It's completely normal and
expected situation.

For #28


ozerovandrei:
@jtopjian it's a really tiny fix that only deletes a `required` tag. But it's really important 😃

jtopjian:
Makes sense :)

LGTM!

PR #30, opened 2018-04-29
set AuthOptions.AllowReauth by default in clientconfig methods

majewsky:
From my experience, this is what most clients want, esp. long-running ones.


jtopjian:
Thanks for pointing out the lack of reauth support here. The reason it isn't currently supported is because (unless I'm mistaken) there's no equivalent to it in the Python libraries.

We can add support for it here (by way of a `clouds.yaml` setting, `clientconfig.AuthInfo` field, and environment variable), but I would prefer not to hard-code the value. IMO, hard-coding the value will run certain pieces of Gophercloud code to run which have been frequently changed and discussed recently. For the class of users who don't require reauthentication, this might cause them to see bugs or behavior they would have otherwise not seen.

majewsky:
Fair enough.

Issue #48, opened 2018-06-05
clientconfig.NewServiceClient should take a RegionName


rickard-von-essen:
It should be possible to pass in `REGION_NAME` to [`clientconfig.NewServiceClient`](https://github.com/gophercloud/utils/blob/master/openstack/clientconfig/requests.go#L432). 

I suggest adding an argument:

```golang
// NewServiceClient is a convenience function to get a new service client.
func NewServiceClient(service, region string, opts *ClientOpts) (*gophercloud.ServiceClient, error) {
```

And just wrap the [region part](https://github.com/gophercloud/utils/blob/master/openstack/clientconfig/requests.go#L474-L482):

```golang
	// Determine the region to use.
	if region == "" {
	        if v := cloud.RegionName; v != "" {
		        region = cloud.RegionName
	        }

	        if v := os.Getenv(envPrefix + "REGION_NAME"); v != "" {
		         region = v
	        }
        }
```

rickard-von-essen:
I don't have a strong opinion, but I think I would prefer `ClientOpts` since it's not regarding auth.

jtopjian:
@rickard-von-essen Yes, good point.

I've opened #49 to add this. Let me know if this looks good to you :)

jtopjian:
Implemented in #49

PR #49, opened 2018-06-06
clientconfig: add RegionName to ClientOpts

jtopjian:
For #48

### DIFF
--- a/openstack/clientconfig/requests.go
+++ b/openstack/clientconfig/requests.go
@@ -41,6 +41,11 @@ type ClientOpts struct {
 	// AuthInfo defines the authentication information needed to
 	// authenticate to a cloud when clouds.yaml isn't used.
 	AuthInfo *AuthInfo
+
+	// RegionName is the region to create a Service Client in.
+	// This will override a region in clouds.yaml or can be used
+	// when authenticating directly with AuthInfo.
+	RegionName string
 }
 
 // LoadYAML will load a clouds.yaml file and return the full config.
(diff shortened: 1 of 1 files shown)


rickard-von-essen:
I prefer if `opts` has the highest priority.

jtopjian:
You mean over environment variables? Environment variables always have the final say.

rickard-von-essen:
I'm working on Packer, I see you have worked on the OpenStack Terraform provider. In both of them application configuration overrides `cloud.yaml` and environment variables.

jtopjian:
> In both of them application configuration overrides cloud.yaml and environment variables.

Unless the application and the application's environment variables are configured to match this `clientconfig` package. 

If Packer is using different environment variables which would conflict with `NewServiceClient`, then I would say `NewServiceClient` isn't suitable for Packer and a custom `NewServiceClient`-ish function local to Packer might be better.

Even though Terraform's environment variables match this package's environment variables, it still does not use `NewServiceClient` as there are too many other finer details it needs to account for. All of Terraform's [`config.go`](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/config.go) file can be thought of as one Terraform-specific `NewServiceClient` function.

rickard-von-essen:
Ok, I thought this was a generalisation of Terraforms `config.go`. But if it's not I think I'll just copy over Terraforms over to Packer and m.m.

jtopjian:
Ah, understood.

This `clientconfig` package as a whole is an attempt to mirror OpenStack's `os-client-config` library, which provides various convenience features and functions to authenticate. There are a bunch of functions within `clientconfig` which can be used together or individually, depending on what works best for the situation at hand. 

It's always possible that certain situations won't be able to use anything in `clientconfig`. In that case, the app developer would have to fall back to `gophercloud/gophercloud` directly.

Terraform does not use the `NewServiceClient` function, but uses `clientconfig` in two places: https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/config.go#L73-L106

It then goes on to configure the rest of the Terraform session in further detail.

Terraform creates a hard separation between someone using a `clouds.yaml` file and someone authenticating with options explicitly set in Terraform. A combination of the two doesn't happen since, as you've raised, precedence becomes very difficult.

On the topic of regions specifically, Terraform handles this in a very unique way. It's possible to set a region at the provider-level and within _each_ resource. 

If you have any questions about what Terraform's doing, please let me know. It's an accumulation of 4 years of work, so there's a lot of functionality accounted for.

But coming back to this PR: I think your request to have `RegionName` within `NewServiceClient` is still very valid. I think `NewServiceClient` will work for 80% of the applications written with Gophercloud. Larger applications, such as Terraform and Packer, will have unique needs outside of it, though.

Issue #45, opened 2018-05-29
clientconfig: Add support for SSL setting to disable verification


haxorof:
In the OpenStack documentation there are some settings which can be made to disable certificate verification. https://docs.openstack.org/os-client-config/latest/user/configuration.html#ssl-settings

It would be good to start with the disable verification switch which is same as setting environment variable OS_INSECURE.

Example below:
```
# clouds.yaml 
clouds:
  insecure: 
    auth: 
      ...
      verify: False
```

This can later be used in Terraform to avoid specifying the insecure config on the OpenStack provider

Cheers!

jtopjian:
If it's a valid `clouds.yaml` feature, then it sounds good to me :)

Two things to note for whoever wants to add this:

1. Similar to https://github.com/gophercloud/gophercloud, we'll need to reference where in the Python libraries this is defined. I've found hunting for clientconfig stuff more difficult than other OpenStack features. There are 3-4 different libraries to look at (os-client-config, osc-lib, openstacksdk, python-openstackclient).

2. This feature should _not_ do any type of http client configuration. That needs to be done at the application-level. Using Terraform as the example, Terraform will still do the TLS configuration [here](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/config.go#L116-L149) but it'll check the `clouds.yaml` configuration (in addition to the Terraform schema).

haxorof:
Python reference:
https://github.com/openstack/os-client-config/blob/stable/queens/os_client_config/cloud_config.py#L114

jtopjian:
Implemented in #50

PR #50, opened 2018-06-06
clientconfig: Add support for SSL setting to disable verification

haxorof:
For #45

@jtopjian : Please review when you got time. I was a bit unsure here what shall be done here. For example in Terraform Cloud struct is fetched [here](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/config.go#L76) then it is easy to just fetch the values directly. Or shall some function be introduced?

### DIFF
--- a/openstack/clientconfig/results.go
+++ b/openstack/clientconfig/results.go
@@ -19,8 +19,7 @@ type Cloud struct {
 	VolumeAPIVersion   string `yaml:"volume_api_version"`
 
 	// Verify whether or not SSL API requests should be verified.
-	// Note! Type is not set to bool to be able to set correct default value.
-	Verify interface{} `yaml:"verify"`
+	Verify *bool `yaml:"verify"`
 
 	// CACertFile a path to a CA Cert bundle that can be used as part of
 	// verifying SSL API requests.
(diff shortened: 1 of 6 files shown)


openstack/clientconfig/results.go
```diff
@@ -17,9 +17,25 @@ type Cloud struct {
 	// API Version overrides.
 	IdentityAPIVersion string `yaml:"identity_api_version"`
 	VolumeAPIVersion   string `yaml:"volume_api_version"`
+
+	// Verify whether or not SSL API requests should be verified.
+	// Note! Type is not set to bool to be able to set correct default value.
+	Verify interface{} `yaml:"verify"`
```
jtopjian: We can change this to `*bool` and the YAML package will account for it accordingly.

This way we can check for nil, true, and false.

openstack/clientconfig/testing/fixtures.go
```diff
@@ -14,6 +14,7 @@ var HawaiiCloudYAML = clientconfig.Cloud{
 		ProjectName: "Some Project",
 		DomainName:  "default",
 	},
+	Verify: true,
```
jtopjian: At the top of this file, do:

```go
var iTrue bool = true
var iFalse bool = false
```

And then set:

```go
Verify: &iTrue
Verify: &iFalse
```

where appropriate.

haxorof:
Thanks @jtopjian! I will fix this straight away. Should have thought of `*bool` but must have been sleeping 😄

haxorof:
Now it is hopefully corrected.

jtopjian:
@haxorof Looks good to me. Thank you for implementing this :)

Issue #42, opened 2018-05-27
Gnocchi: full archive policies support


ozerovandrei:
Currently this repo only supports `List` and `Get` methods for the Gnocchi archive policies.
We need to also add support for the `Create`, `Update` and `Delete` requests.

- [x] Gnocchi archive policy create (#43)
- [x] Gnocchi archive policy update (#44)
- [x] Gnocchi archive policy delete (#52)

Archive policies documentation reference:
https://gnocchi.xyz/rest.html#archive-policy

jtopjian:
@ozerovandrei I have two things to apologize for:

1. I swear I left a comment last week about apologizing for the delay -- I was either imagining that or the comment never posted.
2. I just got a gnocchi environment up and running again, so I will be able to review these shortly. Sorry again for the delay.

ozerovandrei:
@jtopjian there is no need to apologise.
Take your time, after all Gnocchi is not an OpenStack project (at least for now) 😄 

I hope to create additional issues and PRs regarding another Gnocchi objects - `resource types` and `archive policy rules`. So, please do not destroy your Gnocchi environment after this issue!

Also I have another two questions outside of scope of this issue:
- can you get the OpenLab acceptance tests runner working for this repository? (I forgot about acceptance tests by the way, will add them in PRs);
- can I contribute Gnocchi `archive policies`, `resource types` and `archive policy rules` to the terraform-provider-openstack in the future? We discussed Gnocchi status regarding OpenStack in the form of Gophercloud [issue#698](https://github.com/gophercloud/gophercloud/issues/698). But what about the provider? Sorry for off topic. (But if you allow it, acceptance runner in the OpenLab is necessary, I guess).

jtopjian:
@ozerovandrei Thanks for implementing this :)

> can you get the OpenLab acceptance tests runner working for this repository?

I can see if OpenLab can integrate `gophercloud/utils`. It might take some work to get Gnocchi fully integrated, though, as I don't believe there's a devstack plugin for it.

> can I contribute Gnocchi archive policies, resource types and archive policy rules to the terraform-provider-openstack in the future?

I gave this some thought. After working with Gnocchi and revisiting the discussion at https://github.com/gophercloud/gophercloud/issues/698, I think it was a good call not to bundle Gnocchi into Gophercloud. I think Gnocchi is best thought of as a time series database with optional OpenStack support rather than an OpenStack time series database.

[...]

PR #52, opened 2018-06-08
Gnocchi: implement archive policy Delete method

ozerovandrei:
Add the Delete method to the archivepolicies package with tests and documentation example.

For #42

API reference:
https://gnocchi.xyz/rest.html#id5

Code references:
https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/rest/api.py#L299
https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/indexer/sqlalchemy.py#L592

### DIFF
--- a/gnocchi/metric/v1/archivepolicies/requests.go
+++ b/gnocchi/metric/v1/archivepolicies/requests.go
@@ -107,3 +107,14 @@ func Update(client *gophercloud.ServiceClient, archivePolicyName string, opts Up
 
 	return
 }
+
+// Delete accepts a Gnocchi archive policy by its name.
+func Delete(c *gophercloud.ServiceClient, archivePolicyName string) (r DeleteResult) {
+	requestOpts := &gophercloud.RequestOpts{
+		MoreHeaders: map[string]string{
+			"Accept": "application/json, */*",
+		},
+	}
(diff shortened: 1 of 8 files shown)


gnocchi/metric/v1/archivepolicies/requests.go
```diff
@@ -17,3 +17,104 @@ func Get(c *gophercloud.ServiceClient, archivePolicyName string) (r GetResult) {
 	_, r.Err = c.Get(getURL(c, archivePolicyName), &r.Body, nil)
 	return
 }
+
+// CreateOptsBuilder allows to add additional parameters to the Create request.
```
jtopjian:
```go
// CreateOptsBuilder allows extensions to add additional parameters to the Create request.
```

gnocchi/metric/v1/archivepolicies/requests.go
```diff
@@ -17,3 +17,104 @@ func Get(c *gophercloud.ServiceClient, archivePolicyName string) (r GetResult) {
...
+type CreateOptsBuilder interface {
+	ToArchivePolicyCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts specifies parameters of a new subnetpool.
```
jtopjian: `s/subnetpool/Archive Policy/`

Issue #54, opened 2018-06-11
Gnocchi: resource types support


ozerovandrei:
It would be great if this repo supported API for Gnocchi resource types.

Documentation reference: [resource-types](https://gnocchi.xyz/rest.html#resource-types).

- [x] List resource types (#55)
- [x] Get a resource type (#59)
- [x] Create a resource type (#60)
- [x] Update a resource type (#61)
- [x] Delete a resource type (#62)

PR #55, opened 2018-06-16
Gnocchi: implement resource types List method

ozerovandrei:
Add new resourcetypes package with List method.
Add tests and a documentation example.

Code references:
https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/rest/api.py#L953
https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/indexer/sqlalchemy.py#L485
https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/resource_type.py#L264

For #54

### DIFF
--- a/gnocchi/metric/v1/resourcetypes/results.go
+++ b/gnocchi/metric/v1/resourcetypes/results.go
@@ -1,0 +1,145 @@
+package resourcetypes
+
+import (
+	"encoding/json"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// Extract is a function that accepts a result and extracts a Gnocchi resource type.
+func (r commonResult) Extract() (*ResourceType, error) {
+	var s *ResourceType
+	err := r.ExtractInto(&s)
+	return s, err
+}
+
+// GetResult represents the result of a get operation. Call its Extract
+// method to interpret it as a Gnocchi resource type.
+type GetResult struct {
+	commonResult
+}
+
+// ResourceType represents custom Gnocchi resource type.
+type ResourceType struct {
+	// Attributes is a collection of keys and values of different resource types.
+	Attributes []ResourceTypeAttribute `json:"-"`
+
+	// Name is a human-readable resource type identifier.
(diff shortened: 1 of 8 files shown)


gnocchi/metric/v1/resourcetypes/results.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// Name is a human-readable resource type identifier.
+	Name string `json:"name"`
+
+	// State represents current status of a resource type.
+	State string `json:"state"`
+}
+
+// ResourceTypeAttribute represents single attribute of a Gnocchi resource type.
+type ResourceTypeAttribute struct {
```
jtopjian: Can we call this `Attribute`?

ozerovandrei:
@jtopjian Hello.
Attribute was tricky. I haven't started implementing another PR for the #54, because I haven't been sure about current implementation.
Will try to change it according to your request.

jtopjian:
Sounds good. Another option is to leave `Attribute` as a generic `map[string]interface{}`. I guess it depends on if the attributes will be usable or just for informational purposes.

ozerovandrei:
@jtopjian I made changes related to Attribute field. I think that it's better to have at least `Name` and `Type` as separate fields inside the structure because they're common between all attribute types.

jtopjian:
@ozerovandrei This still felt a little strange to me and then I realized it is because we're trying to convert a map of values into an array. 

How about something like this: https://github.com/jtopjian/gophercloud-utils/commit/89b3bbbcc33d84d634a55911e4b44f1f95fd9197

ozerovandrei:
@jtopjian it's because I've tried to do a very strong structure types without any maps from the beginning of that PR. Some Go developer told me to so so once: [#issuecomment-362023884](https://github.com/gophercloud/utils/pull/23#issuecomment-362023884) 😉 

But yeah, I can change this, why not.

jtopjian:
That discussion is still applicable. :)

What I'm proposing is modeling the data in its normal map/object/hash/dict form and not coalescing it into an array/list/slice.

jtopjian:
@ozerovandrei Just to confirm: you're finished with this and it's mergeable?

Issue #56, opened 2018-06-20
Provide a generic REST client


jtopjian:
One request that came up in https://github.com/gophercloud/gophercloud/issues/441 was to have a generic REST client. I think personal opinions might make this type of feature debatable, but given the Python SDK provides this kind of low-level access, it makes sense to provide it here for feature parity.

PR #57, opened 2018-06-20
[WIP] Support for generic REST client

jtopjian:
This commit adds support for a generic REST actions. These actions allow
direct REST communication while still leveraging some of the core
Gophercloud client features.

For #56

### DIFF
--- a/acceptance/openstack/restclient/restclient_test.go
+++ b/acceptance/openstack/restclient/restclient_test.go
@@ -1,0 +1,91 @@
+// +build acceptance restclient
+
+package restclient
+
+import (
+	"testing"
+
+	acc_clients "github.com/gophercloud/gophercloud/acceptance/clients"
+	acc_tools "github.com/gophercloud/gophercloud/acceptance/tools"
+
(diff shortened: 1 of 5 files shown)


jtopjian:
@j-griffith As discussed a few weeks back, here's a first attempt at a "dummy" client. It's able to leverage existing Provider and Service clients, so things like authentication and core HTTP calls are automatically handled. Technically low-level calls are already possible in Gophercloud, but there's a good amount of work involved, especially if you're making multiple calls, so I tried to strike a balance between the core calls and the per-package functionality.

If you can, at your convenience, let me know what you think or if you see any issues.

jtopjian:
Just noticed I can easily add support for URL/query params, too. I'll loop back to this later today.

j-griffith:
Awesome!  I'll try and look at it ASAP

jtopjian:
@j-griffith All sounds good. I'm not in a rush to merge this, so definitely take your time. I'd rather wait a while to hear feedback from yourself or anyone else who notices this PR open :)

Issue #41, opened 2018-05-24
clientconfig: Support for secure.yaml 


simonre:
According to this: https://docs.openstack.org/os-client-config/latest/user/configuration.html Openstack supports a secure.yaml file to specify secrets in a different file. I think it makes sense to include it here as well.

jtopjian:
This coincidentally came up in a Terraform issue today :)

simonre:
@jtopjian I'll add this after making sure that clouds-public.yaml is merged correctly in #40 as I think this is very similar.

haxorof:
References:
https://docs.openstack.org/os-client-config/latest/user/configuration.html#splitting-secrets
https://github.com/openstack/openstacksdk/blob/stable/queens/openstack/config/loader.py

simonre:
@jtopjian @haxorof It seems to me that secure.yaml overrides clouds.yaml and clouds-public.yaml overrides both. This does seem a bit unintuitive to me. Is that assessment correct?

haxorof:
@simonre Yes, basically that is very similar to what I interpret it. Looking at the code it seems to me that the secure.yaml can override everything that is defined in clouds.yaml. And when that is done the clouds-public.yaml only overrides the auth part. 

Merge secure.yaml structure into clouds.yaml structure during construction of OpenStackConfig
https://github.com/openstack/openstacksdk/blob/stable/queens/openstack/config/loader.py#L214-L219

Subsequent calls here will update the auth part with what is defined in clouds-public.yaml if some profile is referenced there:
https://github.com/openstack/openstacksdk/blob/stable/queens/openstack/config/loader.py#L1254

@jtopjian What is your interpretation of this?

jtopjian:
Yes, that sounds correct to me. One other are of code which is relevant is:

https://github.com/openstack/openstacksdk/blob/stable/queens/openstack/config/loader.py#L121-L146

I agree this all seems backwards.

I suppose one way of being absolutely sure is to test this with the three files (`clouds.yaml`, `clouds-public.yaml`, and `secure.yaml`) and run a command like

```
$ openstack server list --os-cloud mycloud
```

haxorof:
Now I did some testing and was a bit surprised over the result. Here is a bullet list of what I found:

- A cloud can be defined in `secure.yaml` without any entry in `clouds.yaml`.
- Not valid to only define the cloud key in `clouds.yaml` even if the rest is defined in `secure.yaml`.
[...]

PR #58, opened 2018-06-25
Support for secure.yaml and public-clouds.yaml

simonre:
For #41 
This PR introduces secure.yaml for clientconfig and also fixes the behaviour of clientconfig when reading public-clouds.yaml as observed by @haxorof in #41

### DIFF
--- a/openstack/clientconfig/requests.go
+++ b/openstack/clientconfig/requests.go
@@ -3,13 +3,13 @@ package clientconfig
 import (
 	"fmt"
 	"os"
+	"reflect"
 	"strings"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
 
 	"gopkg.in/yaml.v2"
-	"reflect"
 )
 
 // AuthType respresents a valid method of authentication.
@@ -161,11 +161,6 @@ func GetCloudFromYAML(opts *ClientOpts) (*Cloud, error) {
 		cloudIsInCloudsYaml = false
 	} else {
 		cloudIsInCloudsYaml = true
-		// Default is to verify SSL API requests
-		if cloud.Verify == nil {
-			iTrue := true
-			cloud.Verify = &iTrue
-		}
 	}
 
 	publicClouds, err := LoadPublicCloudsYAML()
@@ -206,6 +201,12 @@ func GetCloudFromYAML(opts *ClientOpts) (*Cloud, error) {
 				return nil, fmt.Errorf("unable to merge information from clouds.yaml and secure.yaml")
 			}
 		}
+	}
+
+	// Default is to verify SSL API requests
+	if cloud.Verify == nil {
+		iTrue := true
+		cloud.Verify = &iTrue
 	}
 
 	// TODO: this is where reading vendor files should go be considered when not found in
--- a/openstack/clientconfig/requests.go
+++ b/openstack/clientconfig/requests.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack"
 
 	"gopkg.in/yaml.v2"
+	"reflect"
 )
 
 // AuthType respresents a valid method of authentication.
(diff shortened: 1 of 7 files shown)


simonre:
@haxorof @jtopjian Can you please review this?

openstack/clientconfig/requests.go
```diff
@@ -134,14 +155,17 @@ func GetCloudFromYAML(opts *ClientOpts) (*Cloud, error) {
...
-	// Default is to verify SSL API requests
-	if cloud.Verify == nil {
-		iTrue := true
-		cloud.Verify = &iTrue
+		// not an immediate error as it might still be defined in secure.yaml
+		cloudIsInCloudsYaml = false
+	} else {
+		cloudIsInCloudsYaml = true
+		// Default is to verify SSL API requests
+		if cloud.Verify == nil {
```
haxorof: I think this part should be move to the end after all the cloud merges has been done.

openstack/clientconfig/requests.go
```diff
@@ -155,8 +179,33 @@ func GetCloudFromYAML(opts *ClientOpts) (*Cloud, error) {
 		if !ok {
 			return nil, fmt.Errorf("cloud %s does not exist in clouds-public.yaml", profileName)
 		}
+		cloud, err = mergeClouds(cloud, publicCloud)
```
haxorof: If this merge correct? Doesn't this makes the `clouds-public.yaml` to override `clouds.yaml`, should be the other way around.

`secure.yaml` overrides `clouds.yaml` which overrides values coming from `clouds-public.yaml` if `profile` is defined.

simonre: I think it should be correct. The order of arguments is `mergeClouds(override, cloud interface{})`

If you can think of better variable names in the mergeClouds function let me know

haxorof: You're right, I got it backwards there. 

Sometimes the hardest part is to find good names. 😄 I have to think if there are some better names.

openstack/clientconfig/requests.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack"
 
 	"gopkg.in/yaml.v2"
+	"reflect"
```
jtopjian: Since `reflect` is part of the standard library, it should go in between `os` and `strings`. This will fix the Travis error.

jtopjian:
I appreciate the work both of you have done with this. It might take me a few days to find some time to look at it in detail.

openstack/clientconfig/requests.go
simonre: You're right. I though a simple gofmt would suffice but apparently not 😄

jtopjian:
quick update: I haven't had a chance to look at this in-depth. I will have some time this weekend. I apologize for the delay.

simonre:
No worries, thanks for the update

openstack/clientconfig/requests.go
```diff
@@ -155,8 +174,39 @@ func GetCloudFromYAML(opts *ClientOpts) (*Cloud, error) {
...
+		}
+	}
 
-		updateAuthInfo(cloud, &publicCloud)
+	secureClouds, err := LoadSecureCloudsYAML()
+	if err != nil {
+		return nil, fmt.Errorf("unable to load secure.yaml: %s", err)
+	}
+
+	if secureClouds != nil {
```
jtopjian: I might be incorrect, but can this block of code be rewritten as:

```go
// If no entry was found in clouds.yaml, no cloud name was specified,
// and only one secureCloud entry exists, use that as the cloud entry.
if !cloudIsInCloudsYaml && cloudName == "" && len(secureClouds) == 1 {
  for _, v := range secureClouds {
    cloud = &v
  }
}

secureCloud, ok := secureClouds[cloudName]
if !ok && cloud == nil {
  // cloud == nil serves two purposes here:
  // if no entry in clouds.yaml was found and
  // if a single-entry secureCloud wasn't used.
  // At this point, no entry could be determined at all.
  return nil, fmt.Errorf("Could not find cloud %s", cloudName)
}

// If secureCloud has content and it differs from the cloud entry,
// merge the two together.
if !reflect.DeepEqual((Cloud{}), secureCloud) && !reflect.DeepEqual(cloud, secureCloud) {
  cloud, err = mergeClouds(secureCloud, cloud)
  if err != nil {
    return nil, fmt.Errorf("unable to merge information from clouds.yaml and secure.yaml")
  }
}
```

simonre: That order would indeed make more sense I think. I'll change it.
One thing that I just noticed: If clouds.yaml only has a single entry, no cloudName is specified and secure.yaml also only has a single entry, should the entries in clouds.yaml and secure.yaml be merged? Because at the moment, they won't be.

jtopjian: If you can, confirm behavior with the `openstack` tool and let's match what it does. I know when there was 1 x clouds.yaml entry and 1 x vendor entry, I still had to specify a cloud/profile.

haxorof:
LGTMT. @jtopjian not what I can think of right now.

simonre:
@jtopjian I think this is good to go :)

jtopjian:
Sounds good. Thank you both for your work on this -- it was a great collaboration :)

PR #62, opened 2018-07-21
Gnocchi: add Delete method for Resource types

ozerovandrei:
Implement Delete method in the resourcetypes package.
Add unit and acceptance tests with documentation.

Code references:
https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/rest/api.py#L913
https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/indexer/sqlalchemy.py#L540
https://github.com/gnocchixyz/gnocchi/blob/stable/4.2/gnocchi/resource_type.py#L264

For #54


ozerovandrei:
@jtopjian yeah, both of them are ready.

PR #66, opened 2018-08-13
Fix an issue on environments where user.Current() is not implemented.

kkohtaka:
When we call `findAndReadCloudsYAML()` on an environment where user.Current() is not implemented, such as "scratch" Docker container, the function returns an error `unable to get current user: ...`.

However, even on such environments, if we have clouds.yaml on `/etc/openstack/`, the function should load the contents from the file.

### DIFF
--- a/openstack/clientconfig/utils.go
+++ b/openstack/clientconfig/utils.go
@@ -129,15 +129,13 @@ func findAndReadYAML(yamlFile string) ([]byte, error) {
 
 	// unix user config directory: ~/.config/openstack.
 	currentUser, err := user.Current()
-	if err != nil {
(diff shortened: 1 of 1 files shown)


jtopjian:
@kkohtaka Thanks for reporting this. I think the use-case is valid, but I'm not sure if this is the right implementation we want to use.

Can you open an Issue to describe this problem - this way we can separate the issue from the implementation in Github.

kkohtaka:
@jtopjian Thanks for your reply. I opened #67 .

kkohtaka:
Closing since the corresponding issue has been solved.

Issue #69, opened 2018-09-30
clientconfig: Deprioritize Environment Variables


jtopjian:
Currently, environment variables will override any authentication setting that was explicitly defined in the `AuthInfo` struct.

The behaviour should really be the opposite: only use an environment variable if the auth setting is currently empty.

jtopjian:
Fixed in #70

PR #70, opened 2018-09-30
clientconfig: Deprioritize Environment Variables

jtopjian:
For #69 

This commit will make clientconfig only use an environment variable if
an authentication setting was not previously set. This is to fix issues
where an application consuming clientconfig has already determined that
environment variables are not required.

This behavior matches Python-based applications since specifying an
authentication setting on the command-line will take priority of an
environment variable.

/cc @ozerovandrei This will fix the issue seen in https://github.com/terraform-providers/terraform-provider-openstack/issues/434


jtopjian:
@xmik Good catch :)

Please see the latest commit.

xmik:
Now it looks good :)

PR #71, opened 2018-10-26
Don't fail if USER is not defined

flaper87:
There's a way to workaround this issue by setting OS_CLIENT_CONFIG_FILE
to the right path, it's still not ok to fail when a USER is not defined
in the system. This can happen in various scenarios, including running
within a container.

OS_CLIENT_CONFIG_FILE should be a feature and not a workaround for this
problem. If USER is not defined or the `~/.config/openstack` dir
doesn't exist, we should just move on.

### DIFF
--- a/openstack/clientconfig/utils.go
+++ b/openstack/clientconfig/utils.go
@@ -128,16 +128,13 @@ func findAndReadYAML(yamlFile string) ([]byte, error) {
 	}
 
 	// unix user config directory: ~/.config/openstack.
-	currentUser, err := user.Current()
-	if err != nil {
-		return nil, fmt.Errorf("unable to get current user: %s", err)
-	}
-
-	homeDir := currentUser.HomeDir
-	if homeDir != "" {
-		filename := filepath.Join(homeDir, ".config/openstack/"+yamlFile)
-		if ok := fileExists(filename); ok {
-			return ioutil.ReadFile(filename)
+	if currentUser, err := user.Current(); err == nil {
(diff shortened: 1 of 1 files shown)


chaosaffe:
Realistically the library code shouldn't be trying to read env vars or making assumptions about where the config exists, but should be passed in by the code using the library. 
Having an implicit read of an env var and assumptions in general about the execution environment can cause significant pain to the users of the library as the must understand the library's internals to understand the behaviour.

jtopjian:
> Realistically the library code shouldn't be trying to read env vars or making assumptions about where the config exists, but should be passed in by the code using the library.

A few comments about this:

This `clientconfig` package was modeled off of the Python `os-client-config` package which _also_ read in environment variables. I'm not sure it does this because the logic that was in `os-client-config` seems to have been moved around to a few different libraries and I've lost track of what's going on. So there's certainly precedence for this kind of behavior.

Secondly, `gophercloud/utils` is a set of miscellaneous Gophercloud utilities and packages that either _complement_ `gophercloud/gophercloud` or don't directly belong in `gophercloud/gophercloud`. 

If you were to strip out the features of reading in environment variables, home directories, etc from this package, you would end up with `gophercloud/gophercloud`'s `AuthOptions`. If features of this package are giving you grief, that's OK -- thats very similar to any other helper function/utility not matching your specific use-case. Just simply write your own and pass in the logic to `AuthOptions`.

flaper87:
@jtopjian done, thanks for the review!

PR #78, opened 2019-02-12
Enable go modules

radeksimko:
This is in preparation to switching [Terraform OpenStack provider](https://github.com/terraform-providers/terraform-provider-openstack) to go modules (see https://github.com/terraform-providers/terraform-provider-openstack/issues/656).

Enabling go modules throughout the dependency tree of the provider will make the switch and future upgrades easier.

This PR is a result of the following commands (in a clean Go `1.11.5` environment):

```sh
go mod init
go get ./...
go mod tidy
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,14 @@ language: go
 sudo: false
 install:
 - go get -v ./...
-- go get golang.org/x/tools/cmd/goimports
+- GO111MODULE=off go get golang.org/x/tools/cmd/goimports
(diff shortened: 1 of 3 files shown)


radeksimko:
See issue partially related to the test failure: https://github.com/gophercloud/utils/issues/79

radeksimko:
This is now ready for review after replacing the UUID library per https://github.com/gophercloud/utils/issues/79.

PR #81, opened 2019-02-23
Terraform Authentication

jtopjian:
This commit adds initial support for Terraform authentication.

The primary reason for this feature is to decouple the Terraform
Swift Remote State backend support from the Terraform OpenStack
Provider, thus reducing Terraform core's dependency of the OpenStack
Provider.

However, this package can also be used in any other situation where
a user wants to make an OpenStack-based Terraform provider/tool.

For https://github.com/terraform-providers/terraform-provider-openstack/issues/686

### DIFF
--- a/terraform/auth/util.go
+++ b/terraform/auth/util.go
@@ -1,0 +1,78 @@
+package auth
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/mitchellh/go-homedir"
+)
+
+// List of headers that need to be redacted
+var REDACT_HEADERS = []string{"x-auth-token", "x-auth-key", "x-service-token",
+	"x-storage-token", "x-account-meta-temp-url-key", "x-account-meta-temp-url-key-2",
+	"x-container-meta-temp-url-key", "x-container-meta-temp-url-key-2", "set-cookie",
+	"x-subject-token"}
+
+// RedactHeaders processes a headers object, returning a redacted list
+func RedactHeaders(headers http.Header) (processedHeaders []string) {
+	for name, header := range headers {
+		for _, v := range header {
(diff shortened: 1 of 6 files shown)


terraform/auth/util.go
```diff
@@ -0,0 +1,78 @@
...
+var REDACT_HEADERS = []string{"x-auth-token", "x-auth-key", "x-service-token",
+	"x-storage-token", "x-account-meta-temp-url-key", "x-account-meta-temp-url-key-2",
+	"x-container-meta-temp-url-key", "x-container-meta-temp-url-key-2", "set-cookie",
+	"x-subject-token"}
+
+// RedactHeaders processes a headers object, returning a redacted list
+func RedactHeaders(headers http.Header) (processedHeaders []string) {
+	for name, header := range headers {
+		for _, v := range header {
+			if sliceContainsStr(REDACT_HEADERS, name) {
```
kayrus: @jtopjian be aware that the original `sliceContainsStr` was case-insensitive: https://github.com/Unknwon/com/blob/2d12a219ccaf4befccef496869ebf1c957d50709/slice.go#L69

jtopjian: Are there concerns that a header will ever not be sent in all lower case?

kayrus: http/1.x and http/2 send headers in a different way.
http1:
 - X-Something
http2:
 - x-something

Since gophercloud works mainly in http1, headers are sent in `X-Something` format and new `sliceContainsStr` func won't redact headers.

jtopjian: whoops - you're right. I thought we had the case-specific strings embedded in the code, but they're all lower-case. I'll get this fixed.

jtopjian: OK, should be fixed in https://github.com/gophercloud/utils/pull/83

PR #82, opened 2019-03-01
ConfigDrive utilities

stbenjam:
This is an implementation in gophercloud utils that mimics the future support coming to Ironic to automatically build config drives:

  https://storyboard.openstack.org/#!/story/2005083


openstack/baremetal/v1/nodes/configdrive.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"encoding/base64"
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+// A ConfigDrive struct will be used to create a base64-encoded, gzipped ISO9660 image for use with Ironic.
+type ConfigDrive struct {
```
jtopjian: Since a "util" function has a lot more leniency than a core Gophercloud function, feel free to add additional fields here that might be useful. For example, `TemporaryDirectory` or `BuildDirectory` where `TempDir` and `TempFile` can use as the base directory.

stbenjam:
@jtopjian I updated the PR with a BuildDirectory option, I think it should be otherwise good to go

PR #91, opened 2019-10-17
Add advanced openstack client with the logger support

kayrus:
Resolves https://github.com/gophercloud/gophercloud/issues/1748
@ozerovandrei this PR requires your approval as well, since some logic and tests were taken from the https://github.com/terraform-providers/terraform-provider-openstack/blob/493f93252e8071f592f03ad2da3bdf4288dc365f/openstack/types.go


kayrus:
@jtopjian ready for review

jtopjian:
@kayrus Thanks! Overall this looks good.

How about we rename `main.go` to something like `debugger.go` or just `client.go` since `main.go` implies that it is a standalone executable. And maybe add a `doc.go` file to explain how to use it?

kayrus:
@jtopjian done

kayrus:
I have a doubt about the default no-cache option. Probably it is better to disable this header by default.

ozerovandrei:
I also think that adding that header into the generic library it's not really a good idea since standard python-openstackclients don't do that.

It's better to add it in end services like terraform-provider.

client/client.go
```diff
@@ -30,9 +30,9 @@ func (DefaultLogger) Printf(format string, args ...interface{}) {
 type RoundTripper struct {
 	// Default http.RoundTripper
 	Rt http.RoundTripper
-	// If set to true, the HTTP "Cache-Control: no-cache" header will not be added
-	// by default to all API requests
-	DisableNoCacheHeader bool
+	// If set to true, the HTTP "Cache-Control: no-cache" header will be added by
+	// default to all API requests
+	NoCacheHeader bool
```
jtopjian: What are your thoughts on making this more generic, such as a field called `Headers` of `map[string]string`. This will enable people to add additional headers of any type, including disabling caching.

kayrus: done. also added an ability to customize sensitive data masking

client/client.go
```diff
@@ -0,0 +1,322 @@
...
+	headers *http.Header
+	// A pointer to a map of headers to be masked in logger
+	maskHeaders *map[string]struct{}
+	// A custom function to format and mask JSON requests and responses
+	FormatJSON func([]byte) (string, error)
+	// How many times HTTP connection should be retried until giving up
+	MaxRetries int
+	// If Logger is not nil, then RoundTrip method will debug the JSON
+	// requests and responses
+	Logger func(string, ...interface{})
```
jtopjian: Loggers should already be concurrent/thread safe:

* https://golang.org/src/log/log.go?s=2378:2670#L40
* https://github.com/sirupsen/logrus/blob/5bd5a315a50f7f5663063a3f608757fdca721f7f/logger.go#L12-L42

I think having the previous implementation with a printf interface is sufficient.

client/client.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+// List of headers that contain sensitive data.
+var defaultSensitiveHeaders = map[string]struct{}{
```
jtopjian: I've never seen `map[string]struct` used this way. Is there a benefit over doing this than `map[string]interface` and having the values set to `nil`?

client/client.go
```diff
@@ -0,0 +1,322 @@
...
+		if _, ok := (*maskHeaders)[strings.ToLower(header)]; ok {
+			result[headerIdx] = fmt.Sprintf("%s: %s", header, "***")
+		} else {
+			result[headerIdx] = fmt.Sprintf("%s: %s", header, strings.Join(data, " "))
+		}
+		headerIdx++
+	}
+
+	return result
+}
```
jtopjian: This might be a little more readable:

```go
func (rt *RoundTripper) hideSensitiveHeadersData(headers http.Header) []string {
  result := make([]string, len(headers))
  headerIdx := 0

  // this is concurrency safe
  v := rt.maskHeaders
  if v == nil {
    v = &defaultSensitiveHeaders
  }
  maskHeaders := *v

  for header, data := range headers {
    v := strings.ToLower(header)
    if _, ok := maskHeaders[v]; ok {
      result[headerIdx] = fmt.Sprintf("%s: %s", header, "***")
    } else {
      result[headerIdx] = fmt.Sprintf("%s: %s", header, strings.Join(data, " "))
    }
    headerIdx++
  }

  return result
}
```

client/client.go
kayrus: Yes, but I wanted to cover cases, when Logger has been unset in other thread. From my perspective, using interface was too much. Brought it back.

client/client.go
ozerovandrei: I used map because it's implemented over hash table and has O(1) in average so it's simpler and faster to check for keys in it instead of looping over a slice which gives us O(n) time.
https://dave.cheney.net/2018/05/29/how-the-go-runtime-implements-maps-efficiently-without-generics

Using `struct{}` instead of `interface{}` is common when you implement some signalling over channels when you don't need a value.

Empty struct has zero memory consumption vs empty interface. Consider this example:
```
package main

import (
    "fmt"
    "unsafe"
)

func main() {
    var s struct{}
    fmt.Println(unsafe.Sizeof(s))
    var i interface{}
    fmt.Println(unsafe.Sizeof(i))
}
// Prints:
// 0
// 8
```

https://dave.cheney.net/2014/03/25/the-empty-struct

jtopjian: Right, I understand the use of a map. It's the `map[struct]struct` that I've never seen before. I'm fine with keeping it.

client/client.go
kayrus: Today it appeared that interface doesn't allow to specify custom inline logger func.

E.g. I want to customize logger for multiple clients:

```
                    func client(name string) client {
                        ...
                        provider.HTTPClient = http.Client{
                                Transport: &client.RoundTripper{
                                        Rt:     &http.Transport{},
                                        Logger: func(format string, args ...interface{}) {
                                            log.Debugf("[%s] "+format, name, args...)
                                        },
                                },
                        }
```

I didn't find a way to do this with interfaces.

@jtopjian did I miss something?

jtopjian: Correct, custom inline functions are not interfaces. You need to create a struct that implements a `Printf` method. The struct can be as customized as you like, even have no fields, have lots of different methods, just as long as it has a `Printf` method. The `Printf` method can leverage other loggers, possibly combine logging from multiple clients, etc.

An alternative we can do is:

```go
type RoundTripper struct {
  ...
  Logger *io.Writer
}
```

Which would provide the same kind of functionality, but on a lower-level. For example, the noop logger could be removed and instead of [this](https://github.com/gophercloud/utils/blob/master/client/client.go#L163), we could do:

```
log := ioutil.Discard
if rt.Logger != nil {
  log = *rt.Logger
}

log.Printf("OpenStack Request URL: %s %s", request.Method, request.URL)
```

kayrus: Sorry for not being clear. The problem that I need to create a dynamic (inline) interface with a custom method. It looks that it is impossible with interfaces, therefore I'd like to bring back a function.

kayrus: see https://stackoverflow.com/questions/23134439/go-anonymous-method-for-a-structure

jtopjian: I'd like to hear more details about why this can't be solved using standard interfaces and methods. 

Using the sample code you gave above, wouldn't this work?

```go

type MyLogger struct {
  Client string
}

func (mylogger) Printf(format, args... interface{}) {
  log.Printf("[%s] "+format, mylogger.Client, args...)
}

func client(name string) client {
  myLogger := MyLogger{
    Client: name,
  }

  provider.HTTPClient = http.Client{
    Transport: &client.RoundTripper{
      Rt:     &http.Transport{},
      Logger: myLogger,
    },
  }
}
```

I think the area I'm failing to understand is a situation where the input to the logger is going to be so dynamic that it can't be modeled in a struct. In theory, you could even do this:

```go
struct MyLogger struct {
  Foo interface{}
  Bar func()
}

func (ml) Printf(...) {
  // Treat Foo as a slice
  // or a string
  // or something else
  //
  // or call ml.Bar()
}
```

kayrus: Thanks for your hints.I figured out how to achieve what I want:

```go
type myLogger struct {
        Prefix string
}

func (l myLogger) Printf(format string, args ...interface{}) {
        log.Printf("%s [DEBUG] "+format, append([]interface{}{l.Prefix}, args...)...)
}

func clientFunc(name string) clientResult { 
    ...
    provider.HTTPClient = http.Client{
            Transport: &client.RoundTripper{
                    Rt:     &http.Transport{},
                    Logger: &myLogger{Prefix: name},
            },
    }
    ...
}
```

PR #100, opened 2019-11-25
Add "omitempty" tag to marshal clouds right

Fedosin:
Currently when we try to marshal Clouds, it produces a lot of fields with empty strings, which is not correct.

This patch adds "omitempty" tag to all Cloud fields to allow correct marshalling of the structure.

### DIFF
--- a/openstack/clientconfig/results.go
+++ b/openstack/clientconfig/results.go
@@ -16,112 +16,112 @@ type Clouds struct {
 
 // Cloud represents an entry in a clouds.yaml/public-clouds.yaml/secure.yaml file.
 type Cloud struct {
-	Cloud      string        `yaml:"cloud" json:"cloud"`
-	Profile    string        `yaml:"profile" json:"profile"`
-	AuthInfo   *AuthInfo     `yaml:"auth" json:"auth"`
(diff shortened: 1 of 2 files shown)


jtopjian:
@Fedosin Thanks for submitting this.

As well, thank you for including tests! I hate to nit-pick on this, but would it be possible to quickly reformat the fixtures to be:

```go
package testing

import (
  "testing"

  "github.com/gophercloud/utils/openstack/clientconfig"

  th "github.com/gophercloud/gophercloud/testhelper"
  yaml "gopkg.in/yaml.v2"
)

var VirginiaExpected = `clouds:
  virginia:
    auth:
      auth_url: https://va.example.com:5000/v3
      application_credential_id: app-cred-id
      application_credential_secret: secret
    auth_type: v3applicationcredential
    region_name: VA
    verify: true
`

var HawaiiExpected = `clouds:
  hawaii:
    auth:
      auth_url: https://hi.example.com:5000/v3
      username: jdoe
      password: password
      project_name: Some Project
      domain_name: default
    region_name: HNL
    verify: true
`

func TestMarshallCloudToYaml(t *testing.T) {
...
```

Fedosin:
@jtopjian done!

PR #101, opened 2019-11-26
Sync Terraform auth func

kayrus:
This is an attempt to move forward the https://github.com/terraform-providers/terraform-provider-openstack/pull/688


kayrus:
@jtopjian ready for review

PR #118, opened 2020-04-23
Fix a typo for security group path

kayrus:


jtopjian:
ah - good find - thanks!

PR #119, opened 2020-04-28
Make the deletion of network resources more robust

johscheuer:
During the cleanup of a project I ran in several challenges:

```bash
Expected HTTP response code [] when accessing [DELETE https://mycloud/v2.0/floatingips/0083695c-d17b-4ee7-8610-330d25bde798], but got 409 instead
{"NeutronError": {"type": "FipInUseByPortForwarding", "message": "Floating IP 0083695c-d17b-4ee7-8610-330d25bde798 in use by Port Forwarding resources.", "detail": ""}}
Error deleting floating IP: 0083695c-d17b-4ee7-8610-330d25bde798 from project: 7d44d6117b29422382837433bd266141
```

and

```bash
{"NeutronError": {"type": "ServicePortInUse", "message": "Port 184a0435-4b69-4307-9e53-0eb7073ad9f7 cannot be deleted directly via the port API: has device owner network:ha_router_replicated_interface.", "detail": ""}}
```

and

```bash
Expected HTTP response code [] when accessing [DELETE https://mycloud/v2.0/routers/2cb975a9-aebd-4c18-86f9-e56adda6cb9b], but got 409 instead
{"NeutronError": {"type": "RouterInUse", "message": "Router 2cb975a9-aebd-4c18-86f9-e56adda6cb9b still has ports", "detail": ""}}
Error deleting router: router_1 from project: 7d44d6117b29422382837433bd266141
```
and

```
Expected HTTP response code [200] when accessing [PUT https://mycloud/v2.0/routers/dc820e2f-afaf-476a-9337-b3bb86146ce7/remove_router_interface], but got 409 instead
{"NeutronError": {"type": "RouterInterfaceInUseByRoute", "message": "Router interface for subnet 09d9e3b8-e06e-4af7-bcf7-870e483f4372 on router dc820e2f-afaf-476a-9337-b3bb86146ce7 cannot be deleted, as it is required by one or more routes.", "detail": ""}}
Expected HTTP response code [] when accessing [DELETE https://mycloud/v2.0/routers/dc820e2f-afaf-476a-9337-b3bb86146ce7], but got 409 instead
{"NeutronError": {"type": "RouterInUse", "message": "Router dc820e2f-afaf-476a-9337-b3bb86146ce7 still has ports", "detail": ""}}
Error deleting router: idcp_router from project: 3cdd2eab26ff423f873a5ec60d26e146
```

This PR make the purge part a little bit more robust in order to cleanup dependencies.

### DIFF
--- a/openstack/helpers/projectpurge.go
+++ b/openstack/helpers/projectpurge.go
@@ -8,6 +8,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/portforwarding"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
@@ -191,6 +192,27 @@ func clearBlockStorageSnaphosts(projectID string, storageClient *gophercloud.Ser
 	return nil
 }
 
+func clearPortforwarding(networkClient *gophercloud.ServiceClient, fipID string, projectID string) error {
+	allPages, err := portforwarding.List(networkClient, portforwarding.ListOpts{}, fipID).AllPages()
+	if err != nil {
+		return err
+	}
+
+	allPFs, err := portforwarding.ExtractPortForwardings(allPages)
+	if err != nil {
+		return err
+	}
+
+	for _, pf := range allPFs {
+		err := portforwarding.Delete(networkClient, fipID, pf.ID).ExtractErr()
+		if err != nil {
+			return fmt.Errorf("Error deleting floating IP port forwarding: " + pf.ID + " from project: " + projectID)
+		}
+	}
+
+	return nil
+}
+
 func clearNetworkingFloatingIPs(projectID string, networkClient *gophercloud.ServiceClient) error {
 	listOpts := floatingips.ListOpts{
 		TenantID: projectID,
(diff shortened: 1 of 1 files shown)


openstack/helpers/projectpurge.go
```diff
@@ -229,8 +258,14 @@ func clearNetworkingPorts(projectID string, networkClient *gophercloud.ServiceCl
...
 	if len(allPorts) > 0 {
 		for _, port := range allPorts {
+			if port.DeviceOwner == "network:ha_router_replicated_interface" {
+				fmt.Printf("Skipping port: %s with owner: %s\n", port.ID, port.DeviceOwner)
+				continue
+			}
+
 			err = ports.Delete(networkClient, port.ID).ExtractErr()
 			if err != nil {
+				fmt.Println(err)
```
jtopjian, line 268: I think we can remove this?

openstack/helpers/projectpurge.go
```diff
@@ -251,8 +325,30 @@ func clearNetworkingRouters(projectID string, networkClient *gophercloud.Service
...
 	}
+
+	subnets, err := getAllSubnets(projectID, networkClient)
+	if err != nil {
+		return fmt.Errorf("Error fetching subnets project: " + projectID)
+	}
+
 	if len(allRouters) > 0 {
 		for _, router := range allRouters {
+			fmt.Printf("Delete Router: %s\n", router.ID)
```
jtopjian, line 336: I think we can remove this?

openstack/helpers/projectpurge.go
johscheuer: Ups, yes these are some debugging statements. Are there any plans to support a logging framework e.g. it could be useful to see which resources are deleted.

jtopjian: I think if we were to add logging, it would be in the form of a logger interface like here: https://github.com/gophercloud/utils/blob/master/client/client.go

This way we don't have to make a choice of logger that some people would prefer not to use.

openstack/helpers/projectpurge.go
```diff
@@ -229,6 +258,11 @@ func clearNetworkingPorts(projectID string, networkClient *gophercloud.ServiceCl
 	}
 	if len(allPorts) > 0 {
 		for _, port := range allPorts {
+			if port.DeviceOwner == "network:ha_router_replicated_interface" {
+				fmt.Printf("Skipping port: %s with owner: %s\n", port.ID, port.DeviceOwner)
```
jtopjian, line 262: Looks like I missed this one earlier - sorry about that!

PR #127, opened 2020-10-13
Add mutexkv and hashcode packages

dkder3k:
After [some discussion here](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1080) with @ozerovandrei and @kayrus there is an intention to add globally available mutexkv and hashcode packages.

### DIFF
--- a/terraform/auth/config.go
+++ b/terraform/auth/config.go
@@ -13,10 +13,11 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/objectstorage/v1/swauth"
 	osClient "github.com/gophercloud/utils/client"
 	"github.com/gophercloud/utils/openstack/clientconfig"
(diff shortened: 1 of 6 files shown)


dkder3k:
[Here](https://github.com/gophercloud/gophercloud/blob/master/pagination/pager.go#L93) [this test](https://github.com/gophercloud/utils/blob/master/gnocchi/metric/v1/metrics/testing/requests_test.go#L17) falls into infinite loop

ozerovandrei:
It's because of those changes: https://github.com/gophercloud/utils/commit/da0e919a012a7ee88737a21c661504d6dc4d4078

Latest tests run on master merge also failed: https://travis-ci.org/github/gophercloud/utils/builds/728425382?utm_source=github_status&utm_medium=notification

It should be fixed in a separate PR.

ozerovandrei (approved):

PR #130, opened 2020-11-30
compute: add helper to return available AZs

EmilienM:
Add a new function, named ListAvailableAvailabilityZones, that return
a slice of AZs that are actually available.

Note: it uses availabilityzones.List() which doesn't require admin
privileges (comparing to availabilityzones.ListDetail() which does).

### DIFF
--- a/openstack/compute/v2/availabilityzones/utils.go
+++ b/openstack/compute/v2/availabilityzones/utils.go
@@ -1,0 +1,27 @@
+package availabilityzones
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
+)
+
+// ListAvailableAvailabilityZones is a convenience function that return a slice of available Availability Zones.
+func ListAvailableAvailabilityZones(client *gophercloud.ServiceClient) ([]string, error) {
+	var ret []string
+	allPages, err := availabilityzones.List(client).AllPages()
+	if err != nil {
+		return ret, err
+	}
+
+	availabilityZoneInfo, err := availabilityzones.ExtractAvailabilityZones(allPages)
+	if err != nil {
+		return ret, err
+	}
+
+	for _, zoneInfo := range availabilityZoneInfo {
+		if zoneInfo.ZoneState.Available {
+			ret = append(ret, zoneInfo.ZoneName)
+		}
+	}
+	return ret, nil
+}


EmilienM:
WIP for now, I'm currently testing it.

openstack/compute/v2/availabilityzones/utils.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"errors"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
+)
+
+// AZExists is a convienience function that checks if an availability zone exists and is
+// actually available.
+func AZExists(client *gophercloud.ServiceClient, name string) error {
```
pierreprinetti, line 12: In this function, you get allPages for each call. This means that if you have to check multiple AZs on the caller side, calling this function repeatedly calls get expensive.

If I had to choose a function signature for this helper, I'd choose one that returns a map:

```
func AvailabilityZones(client *gophercloud.ServiceClient) ([map]availabilityzones.AvailabilityZone, error)
```

This results in the same number of API calls per function call, however you can use the data multiple times on the caller side:

```
availabilityZones, _ := utils.AvailabilityZones(client) // check the error :)
if _,ok:=availabilityZones(mightExistOrNot);ok {
    # zone exists
}

File: openstack/compute/v2/availabilityzones/utils.go [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+			if zoneInfo.ZoneState.Available {
+				// AZ was found and is available
+				return nil
+			} else {
+				err := "AZ not available: " + zoneInfo.ZoneName
+				return errors.New(err)
+			}
+		}
+	}
+	notExist := "AZ does not exist: " + name
```
pierreprinetti, line 33: On the caller side, how can you tell if the response is "doesn't exist" or if the error is actually caused by a critical failure? You'll probably have to type-assert the error, or worse pattern-match the error string.

If you want to keep the current behaviour, I'd change the return values to `(bool, error)`

openstack/compute/v2/availabilityzones/utils.go
EmilienM: Rather than returning a map where a key is the AZ name and the value is the AZ data, I think returning a list of available AZs is enough.
The reason is that AZ don't have much infos, see:

```
 openstack availability zone list
+-----------+-------------+
| Zone Name | Zone Status |
+-----------+-------------+
| nova      | available   |
+-----------+-------------+
```

So I propose that we create an helper that returns the AZs which are actually available:

```go
// ListAvailableAvailabilityZones is a convienience function that return a slice of
// available Availability Zones.
func ListAvailableAvailabilityZones(client *gophercloud.ServiceClient) ([]string, error) {
	var ret []string
	allPages, err := availabilityzones.List(client).AllPages()
	if err != nil {
		return ret, err
	}

	availabilityZoneInfo, err := availabilityzones.ExtractAvailabilityZones(allPages)
	if err != nil {
		return ret, err
	}

	for _, zoneInfo := range availabilityZoneInfo {
		if zoneInfo.ZoneState.Available {
			ret = append(ret, zoneInfo.ZoneName)
		}
	}
	return ret, nil
}
```

EmilienM:
@jtopjian hey, do you think this one would be fine?

jtopjian:
@EmilienM Yeah, this `gophercloud/utils` repo is pretty flexible with what is accepted. Let me know if/when this is ready to be merged 🙂

EmilienM:
> @EmilienM Yeah, this `gophercloud/utils` repo is pretty flexible with what is accepted. Let me know if/when this is ready to be merged slightly_smiling_face

Thanks ! yes this is ready in my opinion, unless something thinks the function should work otherwise. I'll try to get someone else to review it before EOD but if you don't see any comment feel free to land it.

pierreprinetti:
:+1:  LGTM!

openstack/compute/v2/availabilityzones/utils.go
```diff
@@ -0,0 +1,27 @@
...
+	if err != nil {
+		return ret, err
+	}
+
+	for _, zoneInfo := range availabilityZoneInfo {
+		if zoneInfo.ZoneState.Available {
+			ret = append(ret, zoneInfo.ZoneName)
+		}
+	}
+	return ret, nil
```
EmilienM, line 26: an alternative is to return 3 things:

- list if available AZ
- list of non available AZ
- errors

so the caller can make a difference between an AZ that doesn't exist or one that exists but isn't available.

openstack/compute/v2/availabilityzones/utils.go
```diff
@@ -0,0 +1,27 @@
+package availabilityzones
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
+)
+
+// ListAvailableAvailabilityZones is a convienience function that return a slice available Availability Zones.
```
mandre, line 8: nit: spelling
```suggestion
// ListAvailableAvailabilityZones is a convenience function that return a slice of available Availability Zones.
```

jtopjian (approved):
LGTM - thank you!

EmilienM:
Thanks for your quick review @jtopjian

PR #131, opened 2020-12-02
Consider TLS options when we create a service client

Fedosin:
Now, when we create a service client with gophercloud/utils, TLS options[1] are ignored, which leads to the fact that we can't work with self-signed certificates out of the box. Also it creates inconsistencies with python-openstackclient, where these options matter.

To remove the discrepancy this commit implements full TLS support.
The work is based on the Terraform client[2], and now that code is shared between the clients.

[1] https://github.com/gophercloud/utils/blob/88c6e9e533261350ec733e2e2147a1d3f8eb62de/openstack/clientconfig/results.go#L39-L49
[2] https://github.com/gophercloud/utils/blob/88c6e9e533261350ec733e2e2147a1d3f8eb62de/terraform/auth/config.go#L144-L179

### DIFF
--- a/internal/util.go
+++ b/internal/util.go
@@ -1,8 +1,15 @@
 package internal
 
 import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"os"
 	"reflect"
 	"strings"
+
+	"github.com/mitchellh/go-homedir"
 )
 
 // RemainingKeys will inspect a struct and compare it to a map. Any struct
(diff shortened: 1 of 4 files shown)


openstack/clientconfig/requests.go
```diff
@@ -749,6 +753,44 @@ func NewServiceClient(service string, opts *ClientOpts) (*gophercloud.ServiceCli
...
+		caPath = v
+	}
+
+	// Get the ca-cert-bundle key if there is a value for CA cert.
+	if caPath != "" {
+		caFile, err := ioutil.ReadFile(caPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read CA cert from disk: %s", err)
+		}
+		caCert = string(bytes.TrimSpace(caFile))
```
kayrus, line 775: You're converting bytes to a string, then convert a string to bytes on the line 784. Can you define the `caCert` var as `[]byte` initially?

kayrus (changes requested):
The code is similar to https://github.com/gophercloud/utils/blob/8677e053dcf1f05d0fa0a616094aace04690eb94/terraform/auth/config.go#L164-L174

I suggest to put the logic into one place with an ability to reuse it.

Fedosin:
@kayrus hi! thank you for you comments. I reused the code from the Terraform client as you suggested :)

jtopjian (approved):
This looks good to me. I'll wait for @kayrus to confirm before merging.

internal/util.go
```diff
@@ -32,3 +39,70 @@ func RemainingKeys(s interface{}, m map[string]interface{}) (extras map[string]i
...
+		path, err = homedir.Expand(path)
+		if err != nil {
+			return path, true, err
+		}
+	}
+
+	if _, err := os.Stat(path); err == nil {
+		contents, err := ioutil.ReadFile(path)
+		if err != nil {
+			return string(contents), true, err
```
kayrus, line 102: can you return `[]byte` in this func? this will allow to avoid excessive `string` -> `[]byte` conversion

kayrus (changes requested):
Overall good, except the type conversion

internal/util.go
Fedosin: yes, sure! this conversion is absolutely unnecessary

kayrus (approved):
LGTM, thanks

jtopjian:
@Fedosin Thank you for working on this. @kayrus Thank you for the review.

PR #133, opened 2020-12-07
Add gofmt check to travis CI

Fedosin:
This change allows to keep the code in a better shape automatically.


EmilienM:
:+1:

jtopjian:
@Fedosin Thanks for submitting this

This is kind of already taken care of here: https://github.com/gophercloud/utils/blob/master/script/format

The above script uses [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports) which also does formatting. However, it's lacking support for simplifying code. 

Googling "goimports simply" returns some Github discussion about this (I won't link here as I'm too shy to have this PR referenced in those larger ones 🙂 ). It might be best to refactor `format` so that it calls both `goimports` and `gofmt`.

If you wanted to do this, go for it.

The reason the `format` script might look a little weird is because it is based off of [this one](https://github.com/gophercloud/gophercloud/blob/master/script/format) from Gophercloud. It accounts for a bunch of edge cases we've run into over the years where Go was trying to format wrong files or incorrectly flagging files because of Golang format bugs (for example, [these files](https://github.com/gophercloud/gophercloud/blob/master/script/format#L19-L22) once triggered `go fmt` to report errors even though they were fine. I should check and see if this is still the case).

I recommend setting up `go fmt` to only scan the files from the `find_files` function as well as have the ability to skip files, too. It might only happen on a rare occasion, but having Travis blocked because of a Golang format bug isn't fun.

Let me know if you have any questions.

PR #134, opened 2020-12-11
TLS improvements

Fedosin:
This commit contains two improvements to the existing TLS module:

1. Now the library ignores the fact the CA certificate can be broken (if it can't parse it for some reason, it just leaves an empty string
without a notice). From now we raise an error in this case.

2. It tries to prevent some obvious formatting errors in the certificate like leading and trailing spaces, additional new lines etc.

### DIFF
--- a/internal/util.go
+++ b/internal/util.go
@@ -1,6 +1,7 @@
 package internal
 
 import (
+	"bytes"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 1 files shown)


mandre (approved):
Seems reasonable.

jtopjian (approved):
LGTM - thank you

PR #135, opened 2020-12-19
imageutils: Use the Image service

pierreprinetti:
Use the Image Service instead of the deprecated image endpoint of the
Compute Service.


jtopjian (approved):
LGTM - thank you!

PR #140, opened 2021-01-11
openstack/clientconfig/requests.go: make sure tls config is generated before creating authenticated client

abhinavdahiya:
Previously the openstack.AuthenticatedClient call would create provider and use it to create an authenticated client without the required TLS configuration.

Now the function creates a provider client and updates the HTTPClient before calling Authenticate to setup the client with authentication tokens.

### DIFF
--- a/openstack/clientconfig/requests.go
+++ b/openstack/clientconfig/requests.go
@@ -744,12 +744,6 @@ func NewServiceClient(service string, opts *ClientOpts) (*gophercloud.ServiceCli
 		}
 	}
 
-	// Get a Provider Client
-	pClient, err := AuthenticatedClient(opts)
-	if err != nil {
-		return nil, err
-	}
-
 	// Check if a custom CA cert was provided.
 	// First, check if the CACERT environment variable is set.
 	var caCertPath string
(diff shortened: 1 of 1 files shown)


openstack/clientconfig/requests.go
```diff
@@ -745,7 +745,11 @@ func NewServiceClient(service string, opts *ClientOpts) (*gophercloud.ServiceCli
 	}
 
 	// Get a Provider Client
-	pClient, err := AuthenticatedClient(opts)
+	ao, err := AuthOptions(opts)
+	if err != nil {
+		return nil, err
+	}
+	pClient, err := openstack.NewClient(ao.IdentityEndpoint)
```
mandre, line 752: How about we take the opportunity to move the creation of `pClient` closer to where it's used, on line 803?

Fedosin (approved):
lgtm

openstack/clientconfig/requests.go
abhinavdahiya: @mandre fixed with 9cc2ea9

mandre (commented):
Thanks @abhinavdahiya . LGTM :+1:

jtopjian (approved):
LGTM

Thanks for catching this! Sorry about that.

jtopjian:
Thanks for the reviews, everyone.

It's been almost 4 years since I first started working with TLS and OpenStack authentication. I recall that there were handful of edge cases that were reported, but once it was working with Terraform, I took it for granted and set it aside in my head. From reviewing this implementation with Terraform's, I think this should be good to go, but my memory is a little fuzzy on those edge cases. 

If any are identified, just open a PR and I'm happy to merge.

PR #141, opened 2021-01-24
Terraform: use reference to MutexKV in Config

ozerovandrei:
We don't want to have a copy of the `sync.Locker` in the
`Config` since lock value shold not be copied.

Usually `Config` used as a singleton in different Terraform provider
resources and all of them should use a single `sync.Locker` instead of
different copies.

This commit changes `MutexKV` field of the `Config` to be a reference
instead of a copy and updates `MutexKV` tests to check that `MutexKV` is
used in a proper way.

Currently you can run `govet` linter and find the issue in the following files of the Terraform OpenStack provider:
https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/provider.go#L513
https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/resource_openstack_networking_router_route_v2.go#L58

### DIFF
--- a/terraform/auth/config.go
+++ b/terraform/auth/config.go
@@ -58,7 +58,7 @@ type Config struct {
 	TerraformVersion string
 	SDKVersion       string
 
-	mutexkv.MutexKV
+	*mutexkv.MutexKV
 }
 
 // LoadAndValidate performs the authentication and initial configuration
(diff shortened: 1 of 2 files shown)


jtopjian:
@ozerovandrei Please feel free to merge this when you're ready 🙂

kayrus:
@ozerovandrei good catch, thanks for fixing this.

PR #142, opened 2021-01-28
clouds: support clouds.yml in addition to clouds.yaml

DSpeichert:
While not clearly documented, the native openstacksdk package supports both
extensions. This brings parity with the official Python SDK.

### DIFF
--- a/openstack/clientconfig/utils.go
+++ b/openstack/clientconfig/utils.go
@@ -108,15 +108,27 @@ func FindAndReadCloudsYAML() (string, []byte, error) {
 		}
 	}
 
-	return FindAndReadYAML("clouds.yaml")
+	s, b, err := FindAndReadYAML("clouds.yaml")
+	if s == "" {
+		return FindAndReadYAML("clouds.yml")
+	}
+	return s, b, err
 }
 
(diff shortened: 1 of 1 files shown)


openstack/clientconfig/utils.go
```diff
@@ -99,7 +99,7 @@ func mergeInterfaces(overridingInterface, inferiorInterface interface{}) interfa
 // 4. unix-specific site_config_dir (/etc/openstack/clouds.yaml)
 //
 // If found, the contents of the file is returned.
-func FindAndReadCloudsYAML() (string, []byte, error) {
+func FindAndReadCloudsYAML() (s string, b []byte, err error) {
```
jtopjian, line 102: nit: Can you revert this? I'm not opposed to named return values, but introducing this would break style consistency within the rest of the code.

jtopjian (changes requested):
@DSpeichert Thank you for submitting this - it's appreciated. I've left one comment with respect to the named return values. Please let me know if you have any questions or would like to discuss.

openstack/clientconfig/utils.go
DSpeichert: Done!

jtopjian (approved):
LGTM - thank you!

PR #143, opened 2021-01-28
fixes merging of clouds.yaml & secure.yaml

### LINKED ISSUE #68: Handle slightly more complex `clouds.yaml` formatting
Each region of a cloud can have its own unique `auth_url` key, so that the schema appears like the following:

```yaml
---
clouds:
    mycloud:
        identity_api_version: 3
        endpoint_type: public
        image_format: raw
        auth:
            user_domain_name: 
            project_domain_name: 
            username: 
        regions:
            - name: Region 1
              values:
                auth:
                  auth_url: 
            - name: Region 2
              values:
                auth:
                  auth_url: 
            - name: Region 3
              values:
                auth:
                  auth_url: 
            - name: Region 4
              values:
                auth:
                  auth_url: 
            - name: Region 5
              values:
                auth:
                  auth_url: 
```

See the following issue for background

https://github.com/terraform-providers/terraform-provider-openstack/issues/405

DSpeichert:
Failures due to "json: unsupported type: map[interface {}]interface {}" are
avoided by using jsoniter package. In addition, "complex" clouds.yaml are now
supported with per-region overrides (such as auth_url).

Fixes #68

### DIFF
--- a/openstack/clientconfig/requests.go
+++ b/openstack/clientconfig/requests.go
@@ -299,6 +299,16 @@ func GetCloudFromYAML(opts *ClientOpts) (*Cloud, error) {
 		cloud.Verify = &iTrue
 	}
 
+	// merging per-region value overrides
+	if opts.RegionName != "" {
+		for _, v := range cloud.Regions {
+			if opts.RegionName == v.Name {
+				cloud, err = mergeClouds(v.Values, cloud)
+				break
+			}
+		}
+	}
+
 	// TODO: this is where reading vendor files should go be considered when not found in
 	// clouds-public.yml
 	// https://github.com/openstack/openstacksdk/tree/master/openstack/config/vendors
--- a/openstack/clientconfig/results.go
+++ b/openstack/clientconfig/results.go
@@ -1,5 +1,7 @@
 package clientconfig
 
+import "encoding/json"
+
 // PublicClouds represents a collection of PublicCloud entries in clouds-public.yaml file.
 // The format of the clouds-public.yml is documented at
 // https://docs.openstack.org/python-openstackclient/latest/configuration/
@@ -16,12 +18,12 @@ type Clouds struct {
 
 // Cloud represents an entry in a clouds.yaml/public-clouds.yaml/secure.yaml file.
 type Cloud struct {
-	Cloud      string        `yaml:"cloud,omitempty" json:"cloud,omitempty"`
-	Profile    string        `yaml:"profile,omitempty" json:"profile,omitempty"`
-	AuthInfo   *AuthInfo     `yaml:"auth,omitempty" json:"auth,omitempty"`
-	AuthType   AuthType      `yaml:"auth_type,omitempty" json:"auth_type,omitempty"`
-	RegionName string        `yaml:"region_name,omitempty" json:"region_name,omitempty"`
-	Regions    []interface{} `yaml:"regions,omitempty" json:"regions,omitempty"`
+	Cloud      string    `yaml:"cloud,omitempty" json:"cloud,omitempty"`
+	Profile    string    `yaml:"profile,omitempty" json:"profile,omitempty"`
+	AuthInfo   *AuthInfo `yaml:"auth,omitempty" json:"auth,omitempty"`
+	AuthType   AuthType  `yaml:"auth_type,omitempty" json:"auth_type,omitempty"`
+	RegionName string    `yaml:"region_name,omitempty" json:"region_name,omitempty"`
+	Regions    []Region  `yaml:"regions,omitempty" json:"regions,omitempty"`
 
 	// EndpointType and Interface both specify whether to use the public, internal,
 	// or admin interface of a service. They should be considered synonymous, but
@@ -124,4 +126,52 @@ type AuthInfo struct {
 	// DefaultDomain is the domain ID to fall back on if no other domain has
 	// been specified and a domain is required for scope.
 	DefaultDomain string `yaml:"default_domain,omitempty" json:"default_domain,omitempty"`
+}
+
+// Region represents a region included as part of cloud in clouds.yaml
+// According to Python-based openstacksdk, this can be either a struct (as defined)
+// or a plain string. Custom unmarshallers handle both cases.
+type Region struct {
+	Name   string `yaml:"name,omitempty" json:"name,omitempty"`
+	Values Cloud  `yaml:"values,omitempty" json:"values,omitempty"`
+}
+
+// UnmarshalJSON handles either a plain string acting as the Name property or
+// a struct, mimicking the Python-based openstacksdk.
+func (r *Region) UnmarshalJSON(data []byte) error {
+	var name string
+	if err := json.Unmarshal(data, &name); err == nil {
+		r.Name = name
+		return nil
+	}
+
+	type region Region
+	var tmp region
+	if err := json.Unmarshal(data, &tmp); err != nil {
+		return err
+	}
+	r.Name = tmp.Name
+	r.Values = tmp.Values
+
+	return nil
+}
+
+// UnmarshalYAML handles either a plain string acting as the Name property or
+// a struct, mimicking the Python-based openstacksdk.
+func (r *Region) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var name string
+	if err := unmarshal(&name); err == nil {
+		r.Name = name
+		return nil
+	}
+
+	type region Region
+	var tmp region
+	if err := unmarshal(&tmp); err != nil {
+		return err
+	}
+	r.Name = tmp.Name
+	r.Values = tmp.Values
+
+	return nil
 }
--- a/openstack/clientconfig/testing/clouds.yaml
+++ b/openstack/clientconfig/testing/clouds.yaml
@@ -133,6 +133,18 @@ clouds:
       password: "this should be overwritten by secure.yaml"
       project_name: "Some Project"
     region_name: "PHL"
+  philadelphia_complex:
+    auth:
+      auth_url: "https://phl.example.com:5000/v3"
+      username: "jdoe"
+      password: "password"
+      project_name: "Some Project"
+    regions:
+      - name: PHL1
+        values:
+          auth:
+            auth_url: "https://phl1.example.com:5000/v3"
+      - PHL2
   virginia:
     auth_type: "v3applicationcredential"
     auth:
--- a/openstack/clientconfig/testing/clouds.yaml
+++ b/openstack/clientconfig/testing/clouds.yaml
@@ -152,3 +152,21 @@ clouds:
       application_credential_id: "app-cred-id"
       application_credential_secret: "secret"
     region_name: "VA"
+  disconnected_clouds:
+    auth:
+      username: "jdoe"
+      password: "password"
+      project_name: "Some Project"
+    regions:
+      - name: SOMEWHERE
+        values:
+          auth:
+            auth_url: "https://somewhere.example.com:5000/v3"
+      - name: ANYWHERE
+        values:
+          auth:
+            auth_url: "https://anywhere.example.com:5000/v3"
+      - name: NOWHERE
+        values:
+          auth:
+            auth_url: "https://nowhere.example.com:5000/v3"
--- a/openstack/clientconfig/testing/fixtures.go
+++ b/openstack/clientconfig/testing/fixtures.go
@@ -44,6 +44,46 @@ var PhiladelphiaCloudYAML = clientconfig.Cloud{
 	AuthType: "password",
 }
 
+var PhiladelphiaComplexPhl1CloudYAML = clientconfig.Cloud{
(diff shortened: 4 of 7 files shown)


jtopjian:
> In addition, "complex" clouds.yaml are now supported with per-region overrides (such as auth_url).

This is appreciated - thank you!

> Failures due to "json: unsupported type: map[interface {}]interface {}" are avoided by using jsoniter package. 

Is this an existing issue? What triggers this error?

DSpeichert:
> Is this an existing issue? What triggers this error?

The error is quite masked, as it isn't being printed even in debug mode. It's caused by `json.Marshall` trying to marshal an `interface{}` (in fact, it's always clientconfig.Cloud type), when said interface contains an `interface{}` field (which it did - for `Regions`, which was of `[]interface{}` type) and then that had a map/dict in it (which it did when using "complex clouds" but it did NOT when using a list of strings - such as in unit tests).

Any dict/map that json package Unmarshals onto `interface{}` becomes a `map[ interface{} ]interface{}`, which it can't Marshal back. That limitation only applies to standard json package, which I initially replace with jsoniter to avoid this problem.

I have, however, found a way around that (resulting from understanding what I wrote above). My original patch here created a Region type:
```go
// Region represents a region included as part of cloud in clouds.yaml
type Region struct {
	Name   string      `yaml:"name,omitempty" json:"name,omitempty"`
	Values interface{} `yaml:"values,omitempty" json:"values,omitempty"`
}
```

That type contained the dreaded `interface{}` type for Values, corresponding to (from example in #68):
```yaml
auth:
  auth_url: http:/...
```
Essentially, it was a map that ended up becoming a `map[ interface{} ]interface{}` - impossible to marshal back.

What I did now was to change `Region.Values` to be of `clientconfig.Cloud` type - that is the de facto purpose of this map - to override anything that we recognize as top level structure. This gets rid of any `interface{}` and therefore we can safely use the standard `encoding/json` package.

What we're left with is two problems:

1. I've broken a test (and changed it). Now it becomes impossible to define `regions` as a simple list of strings. I don't think this serves any purpose anyway and it doesn't look to me like the `.Regions` field was ever used (here or in the `terraform-provider-openstack`). Authoritative list of regions lives in the Keystone catalog, storing it in `clouds.yaml` only makes sense when paired with overrides IMHO.
2. It appears that the signature of `mergeClouds(override, cloud interface{}) (*Cloud, error)` could be changed to `mergeClouds(override, cloud *Cloud) (*Cloud, error)`. This could make it more obvious that `interface{}` can go wrong with `encoding/json`.


Btw, running tests locally seems to require actually getting rid of one's own OS_* env vars, or they get found and used in tests.

I could add some tests for the variable override functionality but I wanted to get more feedback first.

jtopjian:
@DSpeichert Thanks for the detailed reply! I think I understand. 

In general, I'm OK with changing the behaviour of the `Regions` field, although I want to make sure that it's fully compatible with the Python libraries. In other words, if the python libraries support both a slice/list of strings and a map/dict of regions, we want to support that here, too.

The reason this is important is because there might be some `clouds.yaml` files in existence that are being used by other applications outside of golang-based ones. Requiring a user to change the `regions` field might break compatibility with those other applications.

It does look like the Python library supports both a list and dict for `regions`: https://github.com/openstack/openstacksdk/blob/4ce235d36dfa3420534ad1b145cc7c5f7438e8e0/openstack/config/loader.py#L410-L424

We might have to implement a custom yaml or json unmarshaller to detect both different types.

DSpeichert:
Custom Unmarshallers added, tests back to original.

It should behave exactly as the Python version you linked. In case of plain strings, Python "expands" it into a `name: <str>`, hence we don't need to worry about a custom Marshaller.

I plan to add a test or two for the overrides.

jtopjian:
@DSpeichert This looks really good. Let me know if you're still working on it or if it's ready for review/merge.

DSpeichert:
I fixed a few typos and squashed it with a more fitting commit message.
Now it's ready for review & merge I think. Thanks!

jtopjian (approved):
LGTM

jtopjian:
@ozerovandrei @kayrus (GitHub is not letting me request a review from you): would you two mind reviewing this since it touches on the Terraform code?

kayrus:
I made a tiny change and expected the tests to pass, but unfortunately they don't. Do I miss something? See details below:

```diff
diff --git a/openstack/clientconfig/testing/clouds.yaml b/openstack/clientconfig/testing/clouds.yaml
index 21a3cbf..5123839 100644
--- a/openstack/clientconfig/testing/clouds.yaml
+++ b/openstack/clientconfig/testing/clouds.yaml
@@ -143,7 +143,7 @@ clouds:
       - name: PHL1
         values:
           auth:
-            auth_url: "https://phl1.example.com:5000/v3"
+            auth_url: "https://phl1.complex.com:5000/v3"
       - PHL2
   virginia:
     auth_type: "v3applicationcredential"
diff --git a/openstack/clientconfig/testing/fixtures.go b/openstack/clientconfig/testing/fixtures.go
index 9105b32..540e5eb 100644
--- a/openstack/clientconfig/testing/fixtures.go
+++ b/openstack/clientconfig/testing/fixtures.go
@@ -54,7 +54,7 @@ var PhiladelphiaComplexPhl1CloudYAML = clientconfig.Cloud{
        Regions: []clientconfig.Region{
                {
                        Name:   "PHL1",
-                       Values: clientconfig.Cloud{AuthInfo: &clientconfig.AuthInfo{AuthURL: "https://phl1.example.com:5000/v3"}},
+                       Values: clientconfig.Cloud{AuthInfo: &clientconfig.AuthInfo{AuthURL: "https://phl1.complex.com:5000/v3"}},
                },
                {
                        Name:   "PHL2",
@@ -74,7 +74,7 @@ var PhiladelphiaComplexPhl2CloudYAML = clientconfig.Cloud{
        Regions: []clientconfig.Region{
                {
                        Name:   "PHL1",
-                       Values: clientconfig.Cloud{AuthInfo: &clientconfig.AuthInfo{AuthURL: "https://phl1.example.com:5000/v3"}},
+                       Values: clientconfig.Cloud{AuthInfo: &clientconfig.AuthInfo{AuthURL: "https://phl1.complex.com:5000/v3"}},
                },
                {
                        Name:   "PHL2",
diff --git a/openstack/clientconfig/testing/requests_test.go b/openstack/clientconfig/testing/requests_test.go
index 73e56f2..a129837 100644
--- a/openstack/clientconfig/testing/requests_test.go
+++ b/openstack/clientconfig/testing/requests_test.go
@@ -65,6 +65,7 @@ func TestGetCloudFromYAML(t *testing.T) {
        for cloud, clientOpts := range allClientOpts {
                actual, err := clientconfig.GetCloudFromYAML(clientOpts)
                th.AssertNoErr(t, err)
+               t.Logf(cloud)
                th.AssertDeepEquals(t, expectedClouds[cloud], actual)
        }
 }
```

```shell
$ go test -v ./openstack/clientconfig/testing/
=== RUN   TestGetCloudFromYAML
    requests_test.go:68: alberta
    requests_test.go:68: philadelphia
    requests_test.go:68: philadelphia_phl1
    convenience.go:236: Failure in requests_test.go, line 69:at .AuthInfo.AuthURL expected "https://phl1.example.com:5000/v3", but got "https://phl1.complex.com:5000/v3"
    convenience.go:35: Failure in requests_test.go, line 69: The structures were different.
--- FAIL: TestGetCloudFromYAML (0.00s)
=== RUN   TestAuthOptionsExplicitCloud
--- PASS: TestAuthOptionsExplicitCloud (0.00s)
=== RUN   TestAuthOptionsOSCLOUD
--- PASS: TestAuthOptionsOSCLOUD (0.00s)
=== RUN   TestAuthOptionsCreationFromCloudsYAML
--- PASS: TestAuthOptionsCreationFromCloudsYAML (0.00s)
=== RUN   TestAuthOptionsCreationFromLegacyCloudsYAML
--- PASS: TestAuthOptionsCreationFromLegacyCloudsYAML (0.00s)
=== RUN   TestAuthOptionsCreationFromClientConfig
--- PASS: TestAuthOptionsCreationFromClientConfig (0.00s)
=== RUN   TestAuthOptionsCreationFromLegacyClientConfig
--- PASS: TestAuthOptionsCreationFromLegacyClientConfig (0.00s)
=== RUN   TestAuthOptionsCreationFromEnv
--- PASS: TestAuthOptionsCreationFromEnv (0.00s)
=== RUN   TestAuthOptionsCreationFromLegacyEnv
--- PASS: TestAuthOptionsCreationFromLegacyEnv (0.00s)
=== RUN   TestGetCloudFromYAMLWithCustomYAMLOpts
requests_test.go:291: Filename: /home/kay/git/sap/utils/openstack/clientconfig/testing/clouds.yaml
requests_test.go:291: Filename: /home/kay/git/sap/utils/openstack/clientconfig/testing/clouds.yaml
--- PASS: TestGetCloudFromYAMLWithCustomYAMLOpts (0.00s)
=== RUN   TestMarshallCloudToYaml
--- PASS: TestMarshallCloudToYaml (0.00s)
FAIL
FAIL    github.com/gophercloud/utils/openstack/clientconfig/testing     0.012s
FAIL
```

DSpeichert:
Hi @kayrus 

In case of PhiladelphiaComplexPhl1CloudYAML fixture, the way it's used makes the top-level `AuthInfo.AuthURL` to be overriden by the Region's `AuthURL`. Therefore, if you modify the AuthUrl for the region, you also need to modify it "above", because clientOps includes `RegionName: "PHL1",`. This is pretty much the whole point of this patch that when a specific region is selected, it's per-region values override the top-level cloud values before being used.

The test does not obviously show that you can define a cloud without auth_url and that it can be provided at the per-region level only. But it does show that for `PhiladelphiaComplexPhl1CloudYAML`, when asked for `PHL1` region, you can overriden auth_url, yet for `PHL2` you do not. They come from a single `clouds.yaml` cloud but need separate fixtures to express the parsed state.

Here's a patch that would pass tests:
```diff
diff --git a/openstack/clientconfig/testing/clouds.yaml b/openstack/clientconfig/testing/clouds.yaml
index 21a3cbf..5123839 100644
--- a/openstack/clientconfig/testing/clouds.yaml
+++ b/openstack/clientconfig/testing/clouds.yaml
@@ -143,7 +143,7 @@ clouds:
       - name: PHL1
         values:
           auth:
-            auth_url: "https://phl1.example.com:5000/v3"
+            auth_url: "https://phl1.complex.com:5000/v3"
       - PHL2
   virginia:
     auth_type: "v3applicationcredential"
diff --git a/openstack/clientconfig/testing/fixtures.go b/openstack/clientconfig/testing/fixtures.go
index 9105b32..3536feb 100644
--- a/openstack/clientconfig/testing/fixtures.go
+++ b/openstack/clientconfig/testing/fixtures.go
@@ -46,7 +46,7 @@ var PhiladelphiaCloudYAML = clientconfig.Cloud{

 var PhiladelphiaComplexPhl1CloudYAML = clientconfig.Cloud{
        AuthInfo: &clientconfig.AuthInfo{
-               AuthURL:     "https://phl1.example.com:5000/v3",
+               AuthURL:     "https://phl1.complex.com:5000/v3",
                Username:    "jdoe",
                Password:    "password",
                ProjectName: "Some Project",
@@ -54,7 +54,7 @@ var PhiladelphiaComplexPhl1CloudYAML = clientconfig.Cloud{
        Regions: []clientconfig.Region{
                {
                        Name:   "PHL1",
-                       Values: clientconfig.Cloud{AuthInfo: &clientconfig.AuthInfo{AuthURL: "https://phl1.example.com:5000/v3"}},
+                       Values: clientconfig.Cloud{AuthInfo: &clientconfig.AuthInfo{AuthURL: "https://phl1.complex.com:5000/v3"}},
                },
                {
                        Name:   "PHL2",
@@ -74,7 +74,7 @@ var PhiladelphiaComplexPhl2CloudYAML = clientconfig.Cloud{
        Regions: []clientconfig.Region{
                {
                        Name:   "PHL1",
-                       Values: clientconfig.Cloud{AuthInfo: &clientconfig.AuthInfo{AuthURL: "https://phl1.example.com:5000/v3"}},
+                       Values: clientconfig.Cloud{AuthInfo: &clientconfig.AuthInfo{AuthURL: "https://phl1.complex.com:5000/v3"}},
                },
                {
                        Name:   "PHL2",
```

kayrus:
@DSpeichert thanks for clarification. Could you please apply the suggested patch? I assume that it reflects more deeper test.

DSpeichert:
@kayrus I believe the existing test already showed the difference (there was `phl.example.com` and `phl1.example.com`). I have added another test that specifically focuses on not having a shared `auth_url` for all regions. I hope that will be fine with you.

kayrus:
@DSpeichert thanks. @jtopjian LGTM

jtopjian:
@kayrus Thanks!

@DSpeichert Thank you very much for your patience and for working on this.

Issue #145, opened 2021-02-16
ClientConfig fails on secure.yaml absence


vooon:
GetCloudFromYAML start to return error `unable to load secure.yaml: no secure.yml file found`.

This condition is wrong:

https://github.com/gophercloud/utils/blob/13abf22518866c8220f3685c811736b252842d1c/openstack/clientconfig/requests.go#L141

Instead of string comparison need to define error (or use standard) + `errors.Is()`.

PR #146, opened 2021-02-16
Fix secure.yml error

jtopjian:
This commit fixes a logic error which was being triggered when
secure.yml didn't exist. This file is optional and so the
warning can be ignored.

For #145


vooon:
@jtopjian i have a bit different approach to fix it at #147.

Note that there similar issue in `clouds-public.yaml` loader.

jtopjian:
@vooon Yes, you're right - thanks.

jtopjian:
Superseded by #147, though if compatibility issues are reported, we'll probably go with this method.

PR #147, opened 2021-02-16
clientconfig: fix file load error checks

### LINKED ISSUE #145: ClientConfig fails on secure.yaml absence

vooon:
In addition wraps underline Errorf's errors.
Requires Go 1.13+

Fix #145.

### DIFF
--- a/openstack/clientconfig/requests.go
+++ b/openstack/clientconfig/requests.go
@@ -1,8 +1,10 @@
 package clientconfig
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
+	"os"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


jtopjian:
@vooon Thanks - this looks like a better approach. I think the requirement of Go 1.13+ is fine.

jtopjian (approved):
LGTM - thank you!

PR #149, opened 2021-03-17
Use default transport settings for our custom TLS HTTP client

### LINKED ISSUE #148: Proxy environment variables ignored when using a self-signed certificate in clouds.yaml
https://github.com/gophercloud/utils/pull/131 introduced a custom HTTP client to make use of the self-signed certificate specified in the `clouds.yaml` file. This change however missed the Proxy setting, ignoring the proxy environment variables.

This means you can't use both a self-signed certificate and a proxy.

We should teach the custom HTTP client to fetch proxy config from the environment.

mandre:
f41c176 introduced a custom HTTP client to make use of the self-signed
certificate specified in the clouds.yaml file. This change however
removed all the settings that came with the DefaultTransport, including
handling of proxy environment variables and default timeouts.

This commit ensures the custom HTTP client inherits its settings from
the default transport.

Requires golang 1.13.

Fixes #148

### DIFF
--- a/openstack/clientconfig/requests.go
+++ b/openstack/clientconfig/requests.go
@@ -817,11 +817,9 @@ func NewServiceClient(service string, opts *ClientOpts) (*gophercloud.ServiceCli
 		pClient.HTTPClient = *opts.HTTPClient
 	} else {
 		// Otherwise create a new HTTP client with the generated TLS config.
-		pClient.HTTPClient = http.Client{
-			Transport: &http.Transport{
-				TLSClientConfig: tlsConfig,
-			},
-		}
+		transport := http.DefaultTransport.(*http.Transport).Clone()
+		transport.TLSClientConfig = tlsConfig
+		pClient.HTTPClient = http.Client{Transport: transport}
 	}
 
 	err = openstack.Authenticate(pClient, *ao)


pierreprinetti:
LGTM! :+1:

openstack/clientconfig/requests.go
```diff
@@ -819,6 +819,7 @@ func NewServiceClient(service string, opts *ClientOpts) (*gophercloud.ServiceCli
 		// Otherwise create a new HTTP client with the generated TLS config.
 		pClient.HTTPClient = http.Client{
 			Transport: &http.Transport{
+				Proxy:           http.ProxyFromEnvironment,
```
mdbooth, line 822: I wondered why we are doing this in an else statement. Are there other http client options which can be passed in which we're ignoring?

Would it work to unconditionally setting pClient.Transport.TLSClientConfig = tlsConfig?

mandre: The conditional is for when the user comes with his own http client, in which case this it what we need to use.

The else branch here is really the normal flow.

mdbooth: What previously set ProxyFromEnvironment?

mandre: I assume the default transport (when you don't specify one) takes care of it.

jtopjian: > The conditional is for when the user comes with his own http client, in which case this it what we need to use.

Exactly correct. It's impossible to support an HTTPClient that will suit everyone's needs, so having the ability to specify a custom one is important.

With this in mind, I'm OK with this PR. `ProxyFromEnvironment` makes sense and helps read in a few environment variables. The reason it wasn't accounted for before was because this HTTPClient logic was taken from Terraform where `ProxyFromEnvironment` isn't used. I understand how adding this statement in a prior PR removed functionality in other applications, though.

However, if there are additional changes to this area of code in the future, I might pass on merging them. My concern would be that the HTTPClient being built here is beginning to get to "app-specific" territory and the custom client might be a better option.

mdbooth: Answering my own question, this was DefaultTransport, which is defined here: https://golang.org/pkg/net/http/

```
var DefaultTransport RoundTripper = &Transport{
    Proxy: ProxyFromEnvironment,
    DialContext: (&net.Dialer{
        Timeout:   30 * time.Second,
        KeepAlive: 30 * time.Second,
    }).DialContext,
    ForceAttemptHTTP2:     true,
    MaxIdleConns:          100,
    IdleConnTimeout:       90 * time.Second,
    TLSHandshakeTimeout:   10 * time.Second,
    ExpectContinueTimeout: 1 * time.Second,
}
```

The implicit behaviour change here goes beyond no longer using the proxy: it looks like we also changed all the various timeouts from their previous values.

Would it be worth, for example, copying DefaultTransport and simply setting Transport.TLSClientConfig in the copy?

jtopjian: Thank you for looking into this. I agree that the best method would be to copy the defaults.

mandre: I'm now making a copy of the DefaultTransport via the Clone() method introduced in golang 1.13.
Are we OK with making golang 1.13 the minimum version?

Otherwise we'll have to copy the DefaultTransport struct here.

openstack/clientconfig/requests.go
```diff
@@ -817,11 +817,9 @@ func NewServiceClient(service string, opts *ClientOpts) (*gophercloud.ServiceCli
 		pClient.HTTPClient = *opts.HTTPClient
 	} else {
 		// Otherwise create a new HTTP client with the generated TLS config.
-		pClient.HTTPClient = http.Client{
-			Transport: &http.Transport{
-				TLSClientConfig: tlsConfig,
-			},
-		}
+		transport := http.DefaultTransport.(*http.Transport).Clone()
```
mdbooth, line 820: Almost identical to my local patch 😀 This is so ugly it's almost like they don't actually want you do to this, though 🤔

mdbooth (approved):
/lgtm

The lack of tests makes me uncomfortable, but as the module doesn't appear to have any test coverage already that's a larger issue than this patch.

openstack/clientconfig/requests.go
mdbooth: This was my local patch:

```go
// Otherwise initialise a new HTTP client based on
// DefaultTransport, and add the generated TLS config.
tlsTransport := *(http.DefaultTransport.(*http.Transport))
tlsTransport.TLSClientConfig = tlsConfig
pClient.HTTPClient.Transport = &tlsTransport
```

Does that do what I think it does (namely a shallow copy)?

kayrus:
@jtopjian give me time till tomorrow. I'll review this.

kayrus (approved):
LGTM

mandre:
Just want to make sure you folks saw this PR makes the minimum go version 1.13 and you're OK with it.

jtopjian (approved):
LGTM - I apologize for the delay.

PR #156, opened 2021-07-20
all: IDsFromName to return multiple results

pierreprinetti:
This patch adds a function in the utils packages that returns all the
IDs of the items with the given name.

This patch leaves the behaviour of the existing `IDFromName` functions
untouched, while it reduces code duplication.

### DIFF
--- a/openstack/blockstorage/extensions/backups/utils.go
+++ b/openstack/blockstorage/extensions/backups/utils.go
@@ -5,38 +5,43 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/backups"
 )
 
-// IDFromName is a convienience function that returns a backups's ID given its name.
+// IDFromName is a convenience function that returns a backup's ID given its
+// name. Errors when the number of items found is not one.
 func IDFromName(client *gophercloud.ServiceClient, name string) (string, error) {
-	count := 0
-	id := ""
-
-	listOpts := backups.ListOpts{
-		Name: name,
-	}
-
-	pages, err := backups.List(client, listOpts).AllPages()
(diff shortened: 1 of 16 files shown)


pierreprinetti:
I deem this patch ready for review (and possible, merge). Thank you for your time!

CC @jtopjian

EmilienM:
very handy!
:+1:

kayrus:
I'd sync the code with other services, compute, image, etc.

jtopjian:
@pierreprinetti

I agree that this is handy - thank you! If it's not too much trouble, @kayrus makes a good point that we should add this for the other resources for consistency. If you don't have the time or desire to do that, just let me know. I'll merge this and then take care of the other resources.

pierreprinetti:
I submit to the will of the people! :grin:

pierreprinetti:
@jtopjian 
If you agree, while I am at it, I am tempted by removing the "if name == name", since we already use ListOpts. WDYT?

https://github.com/gophercloud/utils/blob/94b19323398bb809a11f5f7175fe47ef289ba3b6/openstack/blockstorage/v3/volumes/utils.go#L28

pierreprinetti:
That should be it. I have only updated the latest version per package.

Note that I have removed `if s.Name == name` from the loop when there was a `ListOpts` that filtered by name already.

Please spend a moment checking that everything is alright, because that's a lot of copy-pasta and I need some backup eyes

jtopjian:
I've tested everything but the Manila stuff, since I don't have access to a Manila environment. Short of that, this all looks good to me - thanks so much.

jtopjian (approved):
LGTM - thank you!

PR #162, opened 2021-09-06
Add possibility to set AllowReauth in clientOpts

Lirt:
AllowReauth cannot be set using `clientconfig`

This PR adds possibility to set this feature from AuthInfo in utils/clientconfig.

Signed-off-by: Ondrej Vasko <ondrej.vaskoo@gmail.com>

### DIFF
--- a/openstack/clientconfig/results.go
+++ b/openstack/clientconfig/results.go
@@ -139,7 +139,7 @@ type AuthInfo struct {
 	// the RoundTripper interface and stores the number of failed retries. For an
 	// example of this, see here:
 	// https://github.com/rackspace/rack/blob/1.0.0/auth/clients.go#L311
-	AllowReauth *bool `yaml:"allow_reauth,omitempty" json:"allow_reauth,omitempty"`
+	AllowReauth bool `yaml:"allow_reauth,omitempty" json:"allow_reauth,omitempty"`
 }
 
 // Region represents a region included as part of cloud in clouds.yaml
--- a/openstack/clientconfig/results.go
+++ b/openstack/clientconfig/results.go
@@ -126,6 +126,20 @@ type AuthInfo struct {
 	// DefaultDomain is the domain ID to fall back on if no other domain has
 	// been specified and a domain is required for scope.
 	DefaultDomain string `yaml:"default_domain,omitempty" json:"default_domain,omitempty"`
+
+	// AllowReauth should be set to true if you grant permission for Gophercloud to
+	// cache your credentials in memory, and to allow Gophercloud to attempt to
+	// re-authenticate automatically if/when your token expires.  If you set it to
+	// false, it will not cache these settings, but re-authentication will not be
+	// possible.  This setting defaults to false.
+	//
+	// NOTE: The reauth function will try to re-authenticate endlessly if left
+	// unchecked. The way to limit the number of attempts is to provide a custom
+	// HTTP client to the provider client and provide a transport that implements
+	// the RoundTripper interface and stores the number of failed retries. For an
+	// example of this, see here:
+	// https://github.com/rackspace/rack/blob/1.0.0/auth/clients.go#L311
+	AllowReauth *bool `yaml:"allow_reauth,omitempty" json:"allow_reauth,omitempty"`
 }
 
 // Region represents a region included as part of cloud in clouds.yaml
--- a/openstack/clientconfig/results.go
+++ b/openstack/clientconfig/results.go
@@ -132,13 +132,6 @@ type AuthInfo struct {
 	// re-authenticate automatically if/when your token expires.  If you set it to
 	// false, it will not cache these settings, but re-authentication will not be
 	// possible.  This setting defaults to false.
-	//
-	// NOTE: The reauth function will try to re-authenticate endlessly if left
-	// unchecked. The way to limit the number of attempts is to provide a custom
-	// HTTP client to the provider client and provide a transport that implements
-	// the RoundTripper interface and stores the number of failed retries. For an
-	// example of this, see here:
-	// https://github.com/rackspace/rack/blob/1.0.0/auth/clients.go#L311
 	AllowReauth bool `yaml:"allow_reauth,omitempty" json:"allow_reauth,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


openstack/clientconfig/results.go
```diff
@@ -126,6 +126,20 @@ type AuthInfo struct {
...
+	// re-authenticate automatically if/when your token expires.  If you set it to
+	// false, it will not cache these settings, but re-authentication will not be
+	// possible.  This setting defaults to false.
+	//
+	// NOTE: The reauth function will try to re-authenticate endlessly if left
+	// unchecked. The way to limit the number of attempts is to provide a custom
+	// HTTP client to the provider client and provide a transport that implements
+	// the RoundTripper interface and stores the number of failed retries. For an
+	// example of this, see here:
+	// https://github.com/rackspace/rack/blob/1.0.0/auth/clients.go#L311
```
jtopjian, line 141: This looks to be copied verbatim from https://github.com/gophercloud/gophercloud/blob/513734676e6495f6fec60e7aaf1f86f1ce807428/auth_options.go#L70-L81 which won't work here.

The original comment is out of date (the URL no longer exists). In addition, the text for this comment isn't the right context for the situation in which a developer would be reading this.

Can you please add a descriptive comment about why this option exists in the `clientconfig` package?

jtopjian, line 141: This looks to be copied verbatim from https://github.com/gophercloud/gophercloud/blob/513734676e6495f6fec60e7aaf1f86f1ce807428/auth_options.go#L70-L81 which won't work here.

The original comment is out of date (the URL no longer exists). In addition, the text for this comment isn't the right context for the situation in which a developer would be reading this.

Can you please add a descriptive comment about why this option exists in the clientconfig package?

openstack/clientconfig/results.go
```diff
@@ -126,6 +126,20 @@ type AuthInfo struct {
...
+	AllowReauth *bool `yaml:"allow_reauth,omitempty" json:"allow_reauth,omitempty"`
```
jtopjian, line 142: Why is this `*bool` instead of `bool`? If it was `bool`, I believe you could remove your code in `openstack/clientconfig/requests.go` on line 376-382.

jtopjian (changes requested):
@Lirt Thank you for submitting this. I've left a few comments. Please let me know if you have any questions or need any help.

terraform/auth/config.go
```diff
@@ -150,6 +150,7 @@ func (c *Config) LoadAndValidate() error {
 			ApplicationCredentialID:     c.ApplicationCredentialID,
 			ApplicationCredentialName:   c.ApplicationCredentialName,
 			ApplicationCredentialSecret: c.ApplicationCredentialSecret,
+			AllowReauth:                 &c.AllowReauth,
```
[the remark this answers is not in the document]
Lirt: Okay, done fa6a1fa1fc790b40b1494ed97652f24eafbb7660.

openstack/clientconfig/results.go
Lirt: It's because I wanted to set default value for it in the code. 

I think your point is that `bool` default value is `false` and therefore it would basically set this value to its default (`AllowReauth` should be `false` by default). Do I understand correctly?

I can do it like that, it's just general approach - if the default was `true` I couldn't set it without potentially overwriting user's value.

I am not that experienced with Golang, so I will do as you prefer.

openstack/clientconfig/results.go
Lirt: Yes, it's copied. 

In the code the value is in the end set into `gophercloud.AuthOptions` and that argument to `openstack.AuthenticatedClient`, so I thought that the comment could be same as in gophercloud AuthOptions as it configures the same functionality (I'm just passing argument further).

I can shorten the comment to this (it explains everything):

```go
// AllowReauth should be set to true if you grant permission for Gophercloud to
// cache your credentials in memory, and to allow Gophercloud to attempt to
// re-authenticate automatically if/when your token expires.  If you set it to
// false, it will not cache these settings, but re-authentication will not be
// possible.  This setting defaults to false.
```

Or would you suggest something else?

Lirt:
@jtopjian Hi, thank you for looking into this. 

To provide some background I'm switching from standard gophercloud authentication to clientconfig (to make it easier to load configuration from env/files and so on). I have opened MR https://github.com/Lirt/velero-plugin-for-openstack/pull/27/files#diff-afabe9d12b52566a21946c1aa52e683bf28567c31401312ecbaf042ac2200060L51 and I noticed that compared to clientconfig there are no means to set `AllowReauth`.

kayrus:
@Lirt is it an option for you to call `ao, err := clientconfig.AuthOptions`, then just apply the `ao.AllowReauth = true`?

Lirt:
Hi @kayrus , I can but how would I then run `AuthenticatedClient` function?  I need to provide `ClientOpts` there (https://github.com/gophercloud/utils/blob/bfd010397530bb30f1497cce1e67608f75b58f91/openstack/clientconfig/requests.go#L713).

openstack/clientconfig/results.go
Lirt: I updated to `bool` -- also to avoid complex initialization in multiple places (nil pointers threat)

kayrus:
> I can but how would I then run AuthenticatedClient function?

@Lirt, you can run this function:
https://github.com/kubernetes/cloud-provider-openstack/blob/3801bccc264cb75fd8aa0c84785b9385f234c156/pkg/client/client.go#L312

Lirt:
@kayrus that looks doable. Code looks OK, I will test it.

Lirt:
@kayrus @jtopjian the workaround presented by @kayrus works fine for me.

The question is now, do we close the PR and I will put a comment/link to what approach I made, or do you want to merge this PR into the code?

openstack/clientconfig/results.go
jtopjian: This looks good to me - thank you!

jtopjian:
> The question is now, do we close the PR and I will put a comment/link to what approach I made, or do you want to merge this PR into the code?

If the workaround works, great, but I think we can proceed to merge this. One benefit is that `allow_reauth` could be set in a `clouds.yaml` file (even though it wouldn't be a standard option, but I quickly checked and I think a few of the other options we allow are non-standard, so nothing new there).

openstack/clientconfig/results.go
Lirt: Shortened the commend in 955acf2e8bee58b0659dec6b5e6a6061c7eb8308

Lirt:
@jtopjian Then from my side everything is done.

jtopjian (approved):
LGTM - thank you!

jtopjian:
@Lirt oops - forgot to merge this earlier. Thanks!

PR #169, opened 2022-03-07
Default to blockstorage v3 API

mandre:
The blockstorage v2 API [1] was deprecated in Pike and removed in Xena.
We should make v3 the default API version when none is explicitly
specified.

[1] https://docs.openstack.org/api-ref/block-storage/v2/


mandre:
The unit tests failure is unrelated to my patch - I'm fixing them with https://github.com/gophercloud/utils/pull/170.

PR #170, opened 2022-03-07
Fix client config unit tests

mandre:
Commit b2ceaa8efa963941aa70d34d113d27740a4545e9 broke some the unit
tests by deferring the cleanup of environment variables to when the
function exits. Wrap each test in it's own test function so that cleanup
happens when needed.

### DIFF
--- a/openstack/clientconfig/testing/requests_test.go
+++ b/openstack/clientconfig/testing/requests_test.go
@@ -297,14 +297,16 @@ func TestAuthOptionsCreationFromEnv(t *testing.T) {
 	}
 
 	for cloud, envVars := range allEnvVars {
-		for k, v := range envVars {
-			os.Setenv(k, v)
-			defer os.Unsetenv(k)
-		}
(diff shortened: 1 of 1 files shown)


pierreprinetti:
LGTM!

Note that as an alternative, if you don't want to repeat the teardown, you can leverage the deferred statement by coming up with your own wrapping function:

```Go
	for cloud, envVars := range allEnvVars {
		t.Run(cloud, func(t *testing.T) {
			for k, v := range envVars {
				os.Setenv(k, v)
				defer os.Unsetenv(k)
			}

			actualAuthOpts, err := clientconfig.AuthOptions(nil)
			th.AssertNoErr(t, err)
			th.AssertDeepEquals(t, expectedAuthOpts[cloud], actualAuthOpts)
		})
	}

```

Just make sure to never call `t.Parallel()` within the loop.

mandre:
Yes, thanks for the tip, I believe you're suggestion is actually cleaner as it will create separate test cases per environment. I'll update it.

PR #171, opened 2022-05-17
blockstorage: add helper to return available AZs

stephenfin:
Add a new function, `ListAvailableAvailabilityZones`, that will return a slice
of volume AZs that are available. This is the block storage equivalent of #130.

### DIFF
--- a/openstack/blockstorage/v3/availabilityzones/utils.go
+++ b/openstack/blockstorage/v3/availabilityzones/utils.go
@@ -1,0 +1,35 @@
+package availabilityzones
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/availabilityzones"
+)
+
+// ListAvailableAvailabilityZones is a convenience function that return a slice of available Availability Zones.
+func ListAvailableAvailabilityZones(client *gophercloud.ServiceClient) ([]string, error) {
+	var zones []string
+
+	allPages, err := availabilityzones.List(client).AllPages()
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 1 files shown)


openstack/compute/v2/availabilityzones/utils.go
```diff
@@ -1,27 +1,34 @@
...
 	}
 
 	availabilityZoneInfo, err := availabilityzones.ExtractAvailabilityZones(allPages)
 	if err != nil {
-		return ret, err
+		return zones, err
+	}
+
+	if len(availabilityZoneInfo) == 0 {
+		return nil, fmt.Errorf("could not find an available compute availability zone")
```
pierreprinetti, line 25: Why are you returning an error when there are zero availability zones? I tend to reserve a non-nil error value for failures. Do you think the API would not legitimately return zero AZs?

stephenfin: Yeah, this deserves a comment. Nova will always return at least two AZs: one for internal services and one for everything else. These are typically called `internal` and `nova` but you can configure them with `[availability_zone] internal_service_availability_zone` and `[availability_zone] default_availability_zone`,

stephenfin: Perhaps this is unnecessary logic though and I could leave this to the caller to acknowledge. wdyt?

pierreprinetti: Thanks for the explanation.
It sounds fair to me to let the user deal with the actual Nova response, rather than trying to fix it by raising an error.

stephenfin: Done

PR #183, opened 2023-03-15
terraform: make field AuthOpts public

askarsabyrov:
In order to enable system scoped authorization in terraform-openstack-provider need the field authOpts in Config struct to be public.


askarsabyrov:
@mandre hi, can you have look please? Now I can't use authOpts field in Config struct in terraform-provider-openstack to pass system scope field. Is there any reason to be authOpts field private in struct?
Related PR  in terraform-provider-openstack https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1517

mandre (approved):
Not sure why it was previously private...
LGTM

PR #190, opened 2023-05-22
Provide a ListOpts helper

pierreprinetti:
Here I propose a builder for List queries that allows "OR" clauses.

With `query.New`, it should be easier to list resources based on repeated properties. For example: get information about multiple ports by ID with a single call.

The properties that can be passed with the method `.And(property string, values ...interface{})` are guarded by the fields tagged `q:` in the base filter. For example: when calling `And()` on `query.New(nertworks.ListOpts{})`, the `And` property name must be one of those defined in the `network.ListOpts` filter.

For example:
```Go
package main

import (
	"fmt"
	"os"

	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
	"github.com/gophercloud/gophercloud/pagination"
	"github.com/gophercloud/utils/openstack/clientconfig"
	"github.com/gophercloud/utils/query"
)

func main() {
	networkClient, err := clientconfig.NewServiceClient("network", &clientconfig.ClientOpts{
		Cloud: os.Getenv("OS_CLOUD"),
	})
	if err != nil {
		panic(err)
	}

	q := query.New(networks.ListOpts{
		Status: "ACTIVE",
	}).
		And("name", "port-A", "port-B").
		And("tags", "tag1", "tag2")
		// .And("unexpected-field", "value") <-- this would generate an error when calling `List`

	fmt.Println("Query:", q) // Query: ?name=port-A&name=port-B&status=ACTIVE&tags=tag1&tags=tag2

	if err := networks.List(networkClient, q).EachPage(func(page pagination.Page) (bool, error) {
		n, err := networks.ExtractNetworks(page)
		for _, item := range n {
			fmt.Printf("Found %q\n", item.Name)
		}
		return true, err
	}); err != nil {
		panic(err)
	}
}
```

ListOpts is currently implemented for three Network resources:
* ports
* networks
* subnets

### DIFF
--- a/query/errors.go
+++ b/query/errors.go
@@ -1,0 +1,41 @@
+package query
+
+// As we can't use Go v1.20 in this module yet, this is an adaptation of
+// `errors.Join`.
+// https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/errors/join.go
+
+func joinErrors(errs ...error) error {
+	n := 0
+	for _, err := range errs {
+		if err != nil {
+			n++
+		}
+	}
+	if n == 0 {
+		return nil
+	}
+	e := &joinError{
+		errs: make([]error, 0, n),
+	}
+	for _, err := range errs {
+		if err != nil {
+			e.errs = append(e.errs, err)
+		}
+	}
+	return e
+}
+
+type joinError struct {
+	errs []error
+}
+
+func (e *joinError) Error() string {
+	var b []byte
+	for i, err := range e.errs {
+		if i > 0 {
+			b = append(b, '\n')
+		}
+		b = append(b, err.Error()...)
+	}
+	return string(b)
+}
--- a/query/list.go
+++ b/query/list.go
@@ -1,0 +1,108 @@
+package query
+
+import (
+	"fmt"
+	"net/url"
+	"reflect"
+
+	"github.com/gophercloud/gophercloud"
+)
+
+func New(listOpts interface{}) *ListOpts {
+	availableFields := make(map[string]string)
+	{
+		t := reflect.TypeOf(listOpts)
+		for i := 0; i < t.NumField(); i++ {
+			if tag := t.Field(i).Tag.Get("q"); tag != "" {
+				availableFields[tag] = t.Field(i).Name
+			}
+		}
+	}
+
+	queryString, err := gophercloud.BuildQueryString(listOpts)
+
+	return &ListOpts{
+		allowedFields: availableFields,
+		query:         queryString.Query(),
+		errs:          joinErrors(err),
+	}
+}
+
+// ListOpts can be used to list multiple resources.
+type ListOpts struct {
+	allowedFields map[string]string
+	query         url.Values
+	errs          error
+}
+
+// And adds an arbitrary number of permutations of a single property to filter
+// in. When a single ListOpts is called multiple times with the same property
+// name, the resulting query contains the resulting intersection (AND). Note
+// that how these properties are combined in OpenStack depend on the property.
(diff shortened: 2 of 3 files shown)


pierreprinetti:
It would be nice to be able to easily merge an existing ListOpts object with this

mdbooth:
I would be inclined to add a new struct explicitly for each resource which can use it, alongside the existing ListOpts. Maybe (ports|subnets|networks).ListOptsMulti? Little bit more boilerplate, but at least we wouldn't be accidentally adding support where we shouldn't.

pierreprinetti:
>we wouldn't be accidentally adding support where we shouldn't

Can you give me an example of what you mean?

pierreprinetti:
I have added a bit of reflection. `New` now takes a struct (possibly, one of the existing `ListOpts`) and parses the field names in the `q:"field-name"` tags. `And` methods will reject fields that aren't defined in the base struct.

This gives a minimal (runtime) type check that may or may not add some safety.

pierreprinetti:
also @mdbooth @dulek for review

pierreprinetti:
and @spjmurray if you happen to pass by!

dulek:
Yup, this looks pretty good to me.

mdbooth:
> > we wouldn't be accidentally adding support where we shouldn't
> 
> Can you give me an example of what you mean?

Firstly I don't think we are adding support where we shouldn't because we need to implement the relevant interfaces for each point of use, specifically `ToPortListQuery`, `ToNetworkListQuery`, and `ToSubnetListQuery`. We can create one of these generic ListOpts, but we can't use it anywhere except these 3 specific places.

It feels a bit smelly to have this here. If we can successfully make it generic I would probably put it somewhere like `openstack/networking/v2/networks/utils.go` because it's specifically a neutron thing.

However, this is really an actual part of the API. I feel like it should live in https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/networks/requests.go alongside `ListOpts`. If we just created ListOptsMulti next to ListOpts which is identical except all its fields are slices instead of singletons then it would be compile-time type safe and very simple: basically just boilerplate. We could use something like this code to reduce the boilerplate and avoid implementation errors.

spjmurray:
Ah yes, this is what we need!

Just for discussion purposes we could actually get this working today with something like:

```go
import (
        "fmt"
        "net/url"

        "github.com/gophercloud/gophercloud"
        "github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
)

type MyListOpts struct {
        ports.ListOpts

        Name []string `q:"name"`
}

func BuildCompositeQueryString(opts, base interface{}) (url.Values, error) {
        baseURL, err := gophercloud.BuildQueryString(base)
        if err != nil {
                return nil, err
        }

        optsURL, err := gophercloud.BuildQueryString(opts)
        if err != nil {
                return nil, err
        }

        baseQuery := baseURL.Query()

        for parameter, values := range optsURL.Query() {
                baseQuery.Del(parameter)

                for _, value := range values {
                        baseQuery.Add(parameter, value)
                }
        }

        return baseQuery, nil
}

func (opts MyListOpts) ToPortListQuery() (string, error) {
        values, err := BuildCompositeQueryString(opts, opts.ListOpts)
        return values.Encode(), err
}

func main() {
        opts := MyListOpts{
                ListOpts: ports.ListOpts{
                        Status: "detached",
                        Name: "ignored, hopefully",
                },
                Name: []string{
                        "port1",
                        "port2",
                },
        }

        query, err := opts.ToPortListQuery()
        if err != nil {
                fmt.Println("fail", err)
        }

        fmt.Println(query)
}
```

Kinda ugly though!

PR #193, opened 2023-07-20
Add utility functions for classifying errors

mdbooth:
TODO:
- [ ] Unit tests


pierreprinetti:
Here is a proposal to make sentinel errors useful in the core library: https://github.com/gophercloud/gophercloud/pull/2904
Can you please take a look and see if it covers your case?

PR #196, opened 2023-12-05
openstack/clientconfig: fix dropped errors

alrs:
This picks up two dropped `err` variables in `openstack/clientconfig`.


pierreprinetti (changes requested):
Thank you for this fix.

pierreprinetti:
If you still want to work on this, I think we should have some test coverage of the change you propose.

pierreprinetti (approved):
OK let's merge what we have, which is worthwhile. Thanks for this @alrs

PR #199, opened 2024-02-09
Bump to gophercloud v2.0.0-beta.1 + CI Hygiene

EmilienM:
Bump Gophercloud to v2.0.0-beta.1


pierreprinetti:
What do you think about changing the module name in go.mod to `github.com/gophercloud/utils/v2`? So that the two versions are independently usable

EmilienM:
ping @gophercloud/core

mandre (commented):
Why do we start vendoring?

mandre (commented):
Also, rather than deleting the acceptance tests, shouldn't we try to setup jobs like we did for the gophercloud repo?

EmilienM:
I removed vendoring, but I won't do acceptance, really not worth it IMO.

mandre (approved):
We need a v2 of utils for people to consume gophercloud v2. Let's not block for the deleted acceptance tests.

LGTM.

PR #211, opened 2024-05-29
Replace hashicorp/go-uuid import with gofrs/uuid

awels:
Replace import of hashicorp/go-uuid as its license MPL 2.0 is not incompatible with the license of the project Apache 2.0. Instead use gofrs/uuid which is compatible. It uses an MIT license.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,8 +3,8 @@ module github.com/gophercloud/utils/v2
 go 1.22
 
 require (
+	github.com/gofrs/uuid/v5 v5.2.0
 	github.com/gophercloud/gophercloud/v2 v2.0.0-rc.1
-	github.com/hashicorp/go-uuid v1.0.3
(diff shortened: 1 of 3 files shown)


pierreprinetti:
I am not sure why MPL would be incompatible with this project, but I like the idea of switching to a compliant UUIDv4 generator. What do you think of https://github.com/gofrs/uuid which is also MIT?

awels:
MPL 2.0 is not compatible with Apache 2.0. The project I work on has a license scanner and it flagged the go-uuid dependency as incompatible. I traced it coming from gophercloud as we don't import it ourselves. I am fine with that one. I picked the google one, but if you prefer gofrs that works for me.

pierreprinetti:
Well, thank you then!

awels:
Updated the PR to replace it with gofrs/uuid instead.

pierreprinetti (approved):

PR #219, opened 2024-07-08
Introduce RetriesDefaultWaitTime

marbra-dev:
No retries will be performed in case of a 429 response does not include a `Retry-After` header.
This PR adds a new default wait time, which can be set to perform retries.


pierreprinetti:
I don't think that we should ship generic retry functions in Gophercloud. A retry mechanism can be implemented on the consumer side, possibly by using any of the specialised libraries on the Net.

PR #229, opened 2024-09-26
[terraform]: allow to override service type

### LINKED ISSUE #228: terraform: add an ability to override endpoint service name
Currently there is an ability to override endpoint URL to a static URL. However there is a case to override the endpoint service type to an alias, e.g. replace `object-store` with the `object-store-beta`. This change also must not depend on the original service type, e.g. if the catalog doesn't contain the `object-store`, we still need to replace it with the `object-store-beta` if possible.

There could be 3 scenarios:

There is no `object-store` service in the catalog, but there is `object-store-beta` service and we want to use it:

```
endpoint_overrides = {
  "object-store" = "object-store-beta"
}
```

There is `object-store` service in the catalog, but there is also `object-store-beta` service and we want to use it instead:

```
endpoint_overrides = {
  "object-store" = "object-store-beta"
}
```

There is neither `object-store` nor `object-store-beta` in the catalog, but we just want to override the service URL:

```
endpoint_overrides = {
  "object-store" = "http://localhost/v1/AUTH_679d134152e248118a3c34c4d4900add/"
}
```

kayrus:
Resolves #228

### DIFF
--- a/terraform/auth/config.go
+++ b/terraform/auth/config.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
 
 	"github.com/gophercloud/gophercloud/v2"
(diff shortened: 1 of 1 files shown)


kayrus:
@pierreprinetti @mandre @EmilienM kindly ping

pierreprinetti (approved):
Why do we even have a Terraform directory in utils? What is its purpose?

I have nothing against merging this, but I think we should have a migration plan for moving this code to the Terraform provider, if it's needed there

kayrus:
> Why do we even have a Terraform directory in utils? What is its purpose?

initially this was in the openstack terraform project, but this code is used in other terraform related projects that use non-vanilla openstack services. at some point we decided to move it into the gophercloud/utils to reduce the dependency tree.

pierreprinetti:
Do you think that having all Gophercloud functions accept ServiceClient as an interface ([as proposed here](https://github.com/gophercloud/gophercloud/pull/3203)) would solve this use-case?

kayrus:
In theory yes, but the code base should contain some extra logic. Anyway, I don't think that terraform helper code will be rewritten after introducing the client interface.

pierreprinetti:
>I don't think that terraform helper code will be rewritten after introducing the client interface

Indeed, that's fair. I'll keep my spring-cleaning instincts under control for the time being

Issue #281, opened 2026-05-20
Update to latest gophercloud


rjmercier1:
This repo is used by the terraform openstack provider which we hope to update to use the latest gophercloud.

kayrus:
terraform openstack provider now uses https://github.com/terraform-provider-openstack/utils
see #261

rjmercier1:
root@osadevctl4:~/bob/git/terraform-provider-openstack# find . -type f|xargs egrep github.com/gophercloud/utils/v2
./go.mod:	github.com/gophercloud/utils/v2 v2.0.0-20250710092215-8f6f0255f600
./go.mod:replace github.com/gophercloud/utils/v2 => ../gophercloudutils
./openstack/resource_openstack_compute_instance_v2.go:	flavorsutils "github.com/gophercloud/utils/v2/openstack/compute/v2/flavors"
./openstack/resource_openstack_compute_instance_v2.go:	imagesutils "github.com/gophercloud/utils/v2/openstack/image/v2/images"

both it seems.  

It's using IDFromName for servers and flavors.  Should copy this logic from gophercloud/utils to the terraform utils?

kayrus:
I'm a bit unsure about your PR. It takes a part from the #258, however gophercloud v3 is not yet released. What problem does it solve?

rjmercier1:
Well ultimately this is for the openstack provider issue #2029, which is to add support for port ranges in openstack port forwarding.  I've made the changes to gophercloud but the maintainer is (rightly) requiring the changes be made to latest but the terraform provider is using v2.10.0

terraform-openstack-provider/utils has dependencies on this repo so seems this must be updated as well

kayrus:
there is a main gophercloud branch and v2 branch. all new stuff is merged into the main branch and, if possible, backported to v2. there is no need to adjust dependency to the main gophercloud branch. on your [feature](https://github.com/gophercloud/gophercloud/pull/3756) gets backported, you can just reference the v2 commit and it should work without extra code changes. I'll close all corresponding `Update to latest gophercloud` issues you opened recently.

PR #282, opened 2026-05-20
Update to latest gophercloud

rjmercier1:
For Issue #281


rjmercier1:
The terraform openstack provider uses this repo and we'd like to update it to the latest gophercloud

PR #283, opened 2026-05-29
Add support for OpenStack trusts 

szymon:
This change adds support for trust_id in `AuthInfo` and `OS_TRUST_ID` environment variable.

### DIFF
--- a/openstack/clientconfig/requests.go
+++ b/openstack/clientconfig/requests.go
@@ -655,6 +655,12 @@ func v3auth(cloud *Cloud, opts *ClientOpts) (*gophercloud.AuthOptions, error) {
 		}
 	}
 
+	if cloud.AuthInfo.TrustID == "" {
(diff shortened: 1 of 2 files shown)


openstack/clientconfig/requests.go
```diff
@@ -1038,3 +1046,8 @@ func isApplicationCredential(authInfo *AuthInfo) bool {
 	}
 	return true
 }
+
+// isTrust determines if a trust ID is used to auth.
+func isTrust(authInfo *AuthInfo) bool {
```
kayrus: I don't see the point for this function. It makes more sense to use this condition directly in the 673 line

szymon: removed it

kayrus (approved):

stephenfin (approved):

